PR #14, opened 2023-10-16
poc: Clean up

cjpatton:
* Remove vdpf.py
* Remove ring.py and use concrete fields instead
* Align import formatting with base draft


jimouris:
Everything looks good to me except for one small comment.

Is the end goal only Mastic (:= plasma2 + FLPs)? If yes, then removing the rings is fine. If we end up having both Mastic and plasma2, then we can keep the rings as fields are not needed there. But as you mentioned, for simplicity it makes more sense to just have Fields everywhere.

jimouris (commented):
Looks good!

PR #15, opened 2023-10-16
Clean up the vidpf API and some internals

cjpatton:


jimouris (commented):
What is the purpose of the binder?

jimouris (approved):

cjpatton:
> What is the purpose of the binder?

It's used with the fixed-key AES mode. We pass the VDAF nonce here in order to diversify the key streams across reports.

jimouris (approved):

PR #18, opened 2023-10-16
vidpf: Add parameter for "incremental mode"

cjpatton:
This will be useful for Mastic, where we need to distinguish between doing heavy hitters from the root or aggregation-by-labels at the leaves. Also, I suspect we'll be able to improve communication cost in case Vidpf is used in non-incremental mode.


jimouris (changes requested):
What's the point of the incremental mode?

cjpatton:
> What's the point of the incremental mode?

As described in the commit message: to distinguish between using Mastic for "heaavy hitters" and using Mastic for "aggegation by labels". In the former case, we want to check the proof when we start at the root; for the latter we want to check the proof and only evaluate the leaves.

Also, it should be possible to optimize the size of the public share if we we're not in incremental mode, right?

jimouris:
Oh okay, I see. Maybe we can add it when we need it? It's never read now that's why I asked.

> Also, it should be possible to optimize the size of the public share if we're not in incremental mode, right?

What optimization do you have in mind?

cjpatton:
> Oh okay, I see. Maybe we can add it when we need it? It's never read now that's why I asked.
> 
> > Also, it should be possible to optimize the size of the public share if we're not in incremental mode, right?

I need it for Mastic, which I'm working on now. (All of these PRs today are incremental improvements towards the big one.)


> What optimization do you have in mind?

It's been a while, since I thought about this, so I could be wrong: But isn't it possible to avoid sending a correction word for every level we know we're not going to evaluate (i.e., obtain output shares for)?

jimouris:
> It's been a while, since I thought about this, so I could be wrong: But isn't it possible to avoid sending a correction word for every level we know we're not going to evaluate (i.e., obtain output shares for)?

Hmm, let's discuss about this on Friday, I hadn't considered this case.

jimouris (approved):

cjpatton:
Added a reference to https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/19.

PR #20, opened 2023-10-17
Implement "basic" Mastic

cjpatton:
Stacked on #22.

Add a new module to `poc` that implements the Mastic VDAF. This VDAF is similar to Poplar1 except that instead of aggregating counters, Mastic can aggregate any data type that is compatible with FlpGeneric.

Mastic operates in one of two "modes", depending on how the VIDPF is configured. In "incremental mode", the FLP is checked when evaluating at the first level of the VIDPF tree; in non-incremental mode, the FLP is checked when evaluating the last level of the tree (all intermediate levels are disabled).

When verifying the FLP, the encoded measurement (i.e., `beta`) is computed by evaluating the VIDPF at both child nodes of the root and summing the output shares. By one-hot verifiability, this sum is equal to value claimed by the Client; by path verifiability, checking this value is sufficient each node on the `alpha` path is guaranteed to have the same value.

The code is incomplete but is sufficient to illustrate the basic features we want. Some things to do:

* Add support for circuits with joint randomness

* Specify `PublicShare` and `InputShare`

* Decide whether to bind the query randomness to the VIDPF mode

* Implement message serialization

### DIFF
--- a/poc/vdaf_mastic.py
+++ b/poc/vdaf_mastic.py
@@ -1,0 +1,389 @@
+import sys
+sys.path.append('draft-irtf-cfrg-vdaf/poc')
+
+from common import Unsigned, front, vec_add, vec_sub
+from flp_generic import FlpGeneric
+import hashlib
+from typing import Optional, Union
+from vdaf import Vdaf, test_vdaf
+from vdaf_poplar1 import Poplar1
+from vidpf import Vidpf
+from xof import XofShake128
+
+
+# Domain separation: FLP prove randomness
+USAGE_PROVE_RAND = 0
+
+# Domain separation: FLP Helper proof share
+USAGE_PROOF_SHARE = 1
+
+# Domain separation: FLP query randomness
+USAGE_QUERY_RAND = 2
+
+
+class Mastic(Vdaf):
+    # Operational types and parameters.
+    Field = None # Set by `with_params()`
+    Vidpf = None # Set by `with_params()`
+    Flp = None # Set by `with_params()`
+    Xof = XofShake128
+    ROOT_PROOF = hashlib.sha3_256().digest() # Hash of the empty string
+
+    # Parameters required by `Vdaf`.
+    ID: Unsigned = 0xFFFFFFFF
+    VERIFY_KEY_SIZE = Xof.SEED_SIZE
+    NONCE_SIZE = 16
+    RAND_SIZE = None # Set by `with_params()`
+    SHARES = 2
+    ROUNDS = 1
+
+    # Types required by `Vdaf`
+    AggParam = Poplar1.AggParam
+    PublicShare = None # TODO(cjpatton)
+    InputShare = None # TODO(cjpatton)
+    PrepMessage = None
+
+    @classmethod
+    def shard(cls, measurement, nonce, rand):
+        (vidpf_rand, rand) = front(cls.Vidpf.RAND_SIZE, rand)
+        (prove_rand_seed, rand) = front(cls.Xof.SEED_SIZE, rand)
+        (helper_seed, rand) = front(cls.Xof.SEED_SIZE, rand)
+
+        (alpha, meas) = measurement
+        beta = cls.Flp.encode(meas)
+
+        # Generate VIDPF keys.
+        (init_seed, correction_words, cs_proofs) = \
+            cls.Vidpf.gen(alpha, beta, nonce, vidpf_rand)
+        public_share = (correction_words, cs_proofs)
+
+        # Generate FLP shares.
+        prove_rand = cls.Xof.expand_into_vec(cls.Field,
+            prove_rand_seed,
+            cls.domain_separation_tag(USAGE_PROVE_RAND),
+            b'',
+            cls.Flp.PROVE_RAND_LEN,
+        )
+
+        proof = cls.Flp.prove(beta, prove_rand, [])
+        leader_proof_share = vec_sub(proof,
+                                     cls.helper_proof_share(helper_seed))
+
+        input_shares = [
+            (init_seed[0], leader_proof_share),
+            (init_seed[1], helper_seed),
+        ]
+        return (public_share, input_shares)
+
+    @classmethod
+    def is_valid(cls, agg_param, previous_agg_params):
+        (level, _prefixes) = agg_param
+        if cls.Vidpf.INCREMENTAL_MODE:
+            # The first level evaluated must be `0` and all levels must be distinct.
+            #
+            # TODO(cjpatton) Consider relaxing this check to allow for
+            # "fast-start". Rather than start at the first level, we might want
+            # to start at a later level and more candidate prefixes. For
+            # example, instead of starting at level `0` with all `2`-bit
+            # prefixes, we might start at level `7` with all `8`-bit prefixes.
+            #
+            # The most natural requirement is that we check the FLP the first
+            # time we aggregate the batch. However this can't be enforced
+            # unless we change the structure of the aggregation parameter.
+            # Ideally it's precisely the same as Poplar1 so that we don't have
+            # to change things too much at the DAP level.
+            #
+            # One solution is to make the first level a parameter of the VDAF.
+            # This is probably a good idea anyway, since it's a trade-off the
+            # Aggregators will probably want to agree on anyway.
+            return (
+                (len(previous_agg_params) == 0 and level == 0) or \
+                previous_agg_params[0][0] == 0
+            ) and all(
+                level != other_level
+                for (other_level, _) in previous_agg_params
+            )
+        else:
+            # Only one level may be evaluated and it must be `BITS - 1`.
+            return len(previous_agg_params) == 0 and \
+                level == cls.Vidpf.BITS - 1
+
+
+    @classmethod
+    def prep_init(cls, verify_key, agg_id, agg_param,
+                  nonce, public_share, input_share):
+        (level, prefixes) = agg_param
+        (init_seed, proof_share) = cls.expand_input_share(agg_id, input_share)
+        (correction_words, cs_proofs) = public_share
+
+        # Ensure that candidate prefixes are all unique and appear in
+        # lexicographic order.
+        for i in range(1, len(prefixes)):
+            if prefixes[i-1] >= prefixes[i]:
+                raise ValueError('out of order prefix')
+
+        # Evaluate the VIDPF.
+        (out_share, level_proof) = cls.Vidpf.eval(agg_id,
+                                                  correction_words,
+                                                  init_seed,
+                                                  level,
+                                                  prefixes,
+                                                  cs_proofs,
+                                                  cls.ROOT_PROOF,
+                                                  nonce)
+
+        # Compute the FLP verifier share, if applicable.
+        verifier_share = None
+        if cls.do_range_check(agg_param):
+            # Evaluate the VIDPF at each child of the root node.
+            #
+            # One-hot verifiability: it is sufficient to check the proof over
+            # the sum of the outputs, since VIDPF VIDPF ensures that exactly
+            # one of the children is equal to the encoded `beta` (and the other
+            # is equal to `0`).
+            #
+            # Path verifiability: It is sufficient to check the proof just
+            # once, since the path verifiability property of VIDPF ensures that
+            # the same `beta` is propagated along the entire `alpha` path.
+            #
+            # Implementation note: This invocation of the VIDPF is redundant.
+            # We evaluate at least one (and likely both) of these prefixes
+            # during the main invocation below.
+            (out_share, _level0_proof) = cls.Vidpf.eval(agg_id,
+                                                        correction_words,
+                                                        init_seed,
+                                                        0,
+                                                        [0, 1],
+                                                        cs_proofs,
+                                                        cls.ROOT_PROOF,
+                                                        nonce)
+            meas_share = vec_add(out_share[0], out_share[1])
+
+            query_rand = cls.Xof.expand_into_vec(
+                cls.Flp.Field,
+                verify_key,
+                cls.domain_separation_tag(USAGE_QUERY_RAND),
+                nonce, # TODO(cjpatton) Consider binding to the VIDPF mode
+                cls.Flp.QUERY_RAND_LEN,
+            )
+
+            verifier_share = cls.Flp.query(meas_share,
+                                           proof_share,
+                                           query_rand,
+                                           [], # joint_rand
+                                           cls.SHARES)
+
+        prep_state = []
+        for val_share in out_share:
+            prep_state += cls.Flp.truncate(val_share)
+        prep_share = (level_proof, verifier_share)
+        return (prep_state, prep_share)
+
+    @classmethod
+    def prep_shares_to_prep(cls, agg_param, prep_shares):
+        if len(prep_shares) != 2:
+            raise ValueError('unexpected number of prep shares')
+
+        # Verify the VIDPF output.
+        (level_proof_0, verifier_share_0) = prep_shares[0]
+        (level_proof_1, verifier_share_1) = prep_shares[1]
+        if level_proof_0 != level_proof_1:
+            raise Exception('output vector is not one-hot')
+
+        # Finish verifying the FLP, if applicable.
+        if cls.do_range_check(agg_param):
+            if verifier_share_0 == None or verifier_share_1 == None:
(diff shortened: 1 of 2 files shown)


poc/vdaf_mastic.py
```diff
@@ -0,0 +1,358 @@
...
+            # Evaluate the VIDPF at each child of the root node.
+            #
+            # It is sufficient to check the proof over the sum of the outputs,
+            # since the one-hot verifiability property of VIDPF ensures that
+            # exactly one of the children is equal to the encoded `beta` (and
+            # the other is equal to `0`).
+            #
+            # It is sufficient to check the proof just once, since the path
+            # verifiability property of VIDPF ensures that the same `beta` is
+            # propagated along the entire `alpha` path.
```
cjpatton, line 126: @jimouris please confirm that these "one-hot" and "path" verifiability are properties claimed by VIDPF.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,358 @@
...
+            #
+            # Implementation note: This invocation of the VIDPF is redundant We
+            # evaluate at least one (and likely bot) of these prefixes during
+            # the main invocation below.
+            #
+            # TODO(cjpatton) Evaluate at the root instead, as long as this
+            # doesn't add communication cost.
```
cjpatton, line 133: @jimouris, @hannahdaviscrypto what would you like to do here? In the Overleaf version of the protocol, we evaluate the VIDPF "at the root" (i.e.,at the empty string $\varepsilon$), but this doesn't seem to be possible with the current VIDPF API. (I think Hannah was working on some changes to VIDPF that would have allowed this, but it wasn't compiling and I'm not sure how far she got with it.)

Evaluating at both children and then summing (as we do here) has a bit of CPU overhead. However it also has some advantages:
1. It's simple to specify
2. It has zero communication overhead. (If evaluating at the root, would we need to send more information in the public share so that the Aggregators can reconstruct shares of the output?)
3. It involves at most one more call to `Vidpf.eval_next()` than is necessary. (If `0` is not a prefix of any string in `prefixes`, then it'll be a waste to evaluate it.)

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+from flp_generic import FlpGeneric
+import hashlib
+from typing import Optional, Union
+from vdaf import Vdaf, test_vdaf
+from vdaf_poplar1 import Poplar1
+from vidpf import Vidpf
+from xof import XofShake128
+
+
+USAGE_QUERY_RAND = 0
```
jimouris, line 14: Nit: Can you add a comment to describe what these three are? Also, maybe an `enum` would be cleaner?

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # Generate FLP shares.
+        prove_rand = cls.Xof.expand_into_vec(cls.Field,
+            prove_rand_seed,
+            cls.domain_separation_tag(USAGE_PROVE_RAND),
+            b'',
+            cls.Flp.PROVE_RAND_LEN,
+        )
+
+        proof = cls.Flp.prove(beta, prove_rand, [])
```
jimouris, line 63: What is the last argument [] here?

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+        input_shares = [
+            (init_seed[0], leader_proof_share),
+            (init_seed[1], helper_seed),
+        ]
+        return (public_share, input_shares)
+
+    @classmethod
+    def is_valid(cls, agg_param, previous_agg_params):
+        (level, _prefixes) = agg_param
+        if cls.Vidpf.INCRIMENTAL_MODE:
```
jimouris, line 76: Rename to `INCREMENTAL_MODE`.

poc/vdaf_mastic.py
jimouris: > It is sufficient to check the proof over the sum of the outputs, since the one-hot verifiability property of VIDPF ensures that exactly one of the children is equal to the encoded `beta` (and the other is equal to `0`).

Yeah, that's correct. Just to clarify, this holds if the proofs from both servers are equal.

> It is sufficient to check the proof just once since the path verifiability property of VIDPF ensures that the same `beta` is propagated along the entire `alpha` path.

The path verifiability does not rely on the VIDPF proofs, we get this from the PLASMA parent-children checks. I'd rephrase it to:
```
Path Verifiability: This property ensures that the same `beta` is propagated along the entire `alpha` path. Additionally, for all other evaluations that are not `alpha` (or prefixes of it), we get zero.
```

I'd also add the **One-hot Verifiability:** and **Path Verifiability:** in front of each to make it clear that we rely on these properties.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+    @classmethod
+    def prep_shares_to_prep(cls, agg_param, prep_shares):
+        if len(prep_shares) != 2:
+            raise ValueError('unexpected number of prep shares')
+
+        # Verify the VIDPF output.
+        (level_proof_0, verifier_share_0) = prep_shares[0]
+        (level_proof_1, verifier_share_1) = prep_shares[1]
+        if level_proof_0 != level_proof_1:
+            raise Exception('output vector is not one-hot')
```
jimouris, line 173: How about?
```python
raise Exception('Path Verifiability Error: The output vector is not one-hot.')
```

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+                raise ValueError('prep share with missing verifier share')
+
+            verifier = vec_add(verifier_share_0, verifier_share_1)
+            if not cls.Flp.decide(verifier):
+                raise Exception('programmed measurement is not in range')
+
+        return None
+
+    @classmethod
+    def prep_next(Prio3, prep_state, prep_msg):
```
jimouris, line 187: Prio3?

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+        leader_proof_share = vec_sub(proof,
+                                     cls.helper_proof_share(helper_seed))
+
+        input_shares = [
+            (init_seed[0], leader_proof_share),
+            (init_seed[1], helper_seed),
+        ]
+        return (public_share, input_shares)
+
+    @classmethod
```
jimouris, line 73: Who runs `is_valid()`? I'm a bit confused about what it checks.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,370 @@
...
+            # doesn't add communication cost.
+            (out_share, _level0_proof) = cls.Vidpf.eval(agg_id,
+                                                        correction_words,
+                                                        init_seed,
+                                                        0,
+                                                        [0, 1],
+                                                        cs_proofs,
+                                                        cls.ROOT_PROOF,
+                                                        nonce)
+            meas_share = vec_add(out_share[0], out_share[1])
```
jimouris, line 142: `out_share[0]` and `out_share[1]` are the left and right children, right? Then we add them into `meas_share` and check the FLP, correct? But the full evaluation happens in lines 105-112? Shouldn't we have a check like:

```python
if level == 0:
# check FLP
else:
# do parent-children checks
```

jimouris (changes requested):
Looks good, thanks for putting it together! I added some suggestions, some questions, and some nits.

poc/vdaf_mastic.py
cjpatton: joint randomness (currently empty)

poc/vdaf_mastic.py
cjpatton: can you send a PR with the necessary change to `vidpf`?

poc/vdaf_mastic.py
cjpatton: This isn't Path Verifiability though, right? This is One-hot Verifiability.

poc/vdaf_mastic.py
cjpatton: See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-07#section-5.3

poc/vdaf_mastic.py
cjpatton: Err actually what do you mean by "latter"? If you want to evaluate at $\varepsilon$ (I agree this is cleaner) then I believe we need to change the API. Or am I misunderstanding something?

poc/vdaf_mastic.py
cjpatton: I claim that you always want to do VIDPF verification and only do FLP verification once.

poc/vdaf_mastic.py
jimouris: Correct. We only need 1 FLP to check \epsilon (or left and right) as you did, and then we'll need the parent-children checks to check that this beta value is propagated correctly. Something like: https://github.com/jimouris/mastic/blob/main/src/collect.rs#L246

poc/vdaf_mastic.py
jimouris: I meant that I prefer evaluating the 2 children. VIDPF won't need any changes either.

poc/vdaf_mastic.py
jimouris: Correct, I meant *One-hot* verifiability.

poc/vdaf_mastic.py
jimouris: I might misunderstood something. By VIDPF verification you mean check the proofs? Yes, we should do that at every level. Additionally, we'd need to do the parent-children checks at every level to make sure that the beta value (which we verified is valid in level 0) is propagated correctly down the tree.

jimouris: Exactly. From Overleaf:

> So far, with VIDPF, the servers can verify that there is at most one non-zero output at each level. However, VIDPFs alone are not sufficient to guarantee that the keys are well formed. The servers still need to verify that: a) the non-zero values of the client are
across a single path in the binary tree, and b) the value of the root node is correctly propagated down the VIDPF tree. We formalize these requirements as an additional property: Path Verifiability: blah blah blah

The idea is that we want $y^γ - (y^{γ||0} + y^{γ||1}) = 0$. Then the following should hold:
$y_0^γ - y_0^{γ||0} - y_0^{γ||1} = - y_1^γ + y_1^{γ||0} + y_1^{γ||1}$. So each server computes their part (server 0 computes the left and server 1 computes the right) and they check if it's correct.

cjpatton: Ohhhh that's so beautifully simple. I think you already mentioned this at some point but it didn't bleed through.

To implement this, I think we need to to take @hannahdaviscrypto's changes to `vidpf`. I'll work on this.

jimouris: Thanks :) It might make more sense to do it on the mastic-vdaf level rather than in VIDPF. This might help as well: https://github.com/jimouris/mastic/blob/main/src/collect.rs#L246

jimouris (approved):

poc/vdaf_mastic.py
cjpatton: I ended up extending VIDPF (see top-comment for justification): https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/22

cjpatton:
Rebased (on #22) and squashed.

poc/vdaf_mastic.py
cjpatton: Done.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,389 @@
...
+        (correction_words, cs_proofs) = public_share
+
+        # Ensure that candidate prefixes are all unique and appear in
+        # lexicographic order.
+        for i in range(1, len(prefixes)):
+            if prefixes[i-1] >= prefixes[i]:
+                raise ValueError('out of order prefix')
+
+        # Evaluate the VIDPF.
+        (out_share, level_proof) = cls.Vidpf.eval(agg_id,
```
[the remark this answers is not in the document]
cjpatton: We could eliminate it, but we'd need to make changes to `Vidpf`. In particular, we need to be able to compute our share of `beta`. Perhaps we could have `Vidpf` return `prefix_tree_share`?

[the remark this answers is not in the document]
jimouris: What do you mean, why? Isn't that the `out_share`?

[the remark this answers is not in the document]
cjpatton: To check the FLP we have to compute a share of `beta`. The way we do that here is by evaluating the VIDPF at the children of the root and adding them up.

jimouris (approved):

cjpatton:
Rebased.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,389 @@
...
+        (level, _prefixes) = agg_param
+        if cls.Vidpf.INCREMENTAL_MODE:
+            # The first level evaluated must be `0` and all levels must be distinct.
+            #
+            # TODO(cjpatton) Consider relaxing this check to allow for
+            # "fast-start". Rather than start at the first level, we might want
+            # to start at a later level and more candidate prefixes. For
+            # example, instead of starting at level `0` with all `2`-bit
+            # prefixes, we might start at level `7` with all `8`-bit prefixes.
+            #
```
hannahedavis, line 89: I think it might make more sense to do something like a 'fast-forward' approach rather than a fast-start. Even if we were to start at level 7 in this way, we'd still have to evaluate the entire tree and the FLP at level 0. So we could think about changing the agg_param format to a tuple of agg_params; is_valid would evaluate the same properties on the union of the tuples as it currently does. That way you could just evaluate levels 0 and 7 at once and then move on.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,389 @@
...
+    @classmethod
+    def prep_shares_to_prep(cls, agg_param, prep_shares):
+        if len(prep_shares) != 2:
+            raise ValueError('unexpected number of prep shares')
+
+        # Verify the VIDPF output.
+        (level_proof_0, verifier_share_0) = prep_shares[0]
+        (level_proof_1, verifier_share_1) = prep_shares[1]
+        if level_proof_0 != level_proof_1:
+            raise Exception('output vector is not one-hot')
```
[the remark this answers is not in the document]
cjpatton: Please send a PR to fix since this got merged already!

poc/vdaf_mastic.py
cjpatton: Sounds reasonable. At this stage I don't want to overcomplicate the spec, so I suggest documenting this idea in an issue. You might also consider updating this comment in a PR.

poc/vdaf_mastic.py
```diff
@@ -0,0 +1,389 @@
...
+            # during the main invocation below.
+            (out_share, _level0_proof) = cls.Vidpf.eval(agg_id,
+                                                        correction_words,
+                                                        init_seed,
+                                                        0,
+                                                        [0, 1],
+                                                        cs_proofs,
+                                                        cls.ROOT_PROOF,
+                                                        nonce)
+            meas_share = vec_add(out_share[0], out_share[1])
```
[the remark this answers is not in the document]
cjpatton: Sounds reasonable.

poc/vidpf.py
```diff
@@ -116,42 +121,46 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
...
+        # Compute the path verifier.
+        sha3 = hashlib.sha3_256()
+        for prefix in prefixes:
+            for current_level in range(level):
+                node = prefix >> (level - current_level)
+                y =  prefix_tree_share[(node,        current_level)  ][2]
+                y0 = prefix_tree_share[(node<<1,     current_level+1)][2]
+                y1 = prefix_tree_share[((node<<1)|1, current_level+1)][2]
+                sha3.update(cls.Field.encode_vec(vec_sub(y, vec_add(y0, y1))))
+        path_proof = sha3.digest()
```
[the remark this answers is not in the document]
cjpatton: Right. I attempted to address with an "implementation note" above. Basically you can cache the `prefix_tree_share` across calls to `eval()`. Does this ake sense?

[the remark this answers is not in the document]
cjpatton: https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/main/poc/vidpf.py#L109

PR #22, opened 2023-10-17
vidpf: Add PLASMA-style path verifiability

cjpatton:
As discussed here in https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/20#discussion_r1362349956, for each node in the prefix tree we wish to check that the children sum up to the same value.

Previously we had discussed handling path verifiability at a higher level, in Mastic itself. Indeed, it's possible to use the current VIDPF in a black-box manner and perform the same calculation. However, the resulting algorithm would have asymptotic runtime that is far worse than optimal. Instead, I followed @hannahdaviscrypto's suggestion to modify VIDPF so that it does the path verification itself. The resulting algorithm is as fast as we can hope for, at least asymptotically. (Concretely there is more room for optimization.)

When specifying a complex algorithm like this, it's often useful to write down the asymptotically-optimal version. I think it makes sense to do the same thing here.

This PR also stacks some minor changes to `vidpf` that are technically separate. It may help to review commit-by-commit.

### DIFF
--- a/poc/vidpf.py
+++ b/poc/vidpf.py
@@ -3,8 +3,8 @@
 import sys
 sys.path.append('draft-irtf-cfrg-vdaf/poc')
 
-from common import (ERR_INPUT, format_dst, gen_rand, vec_add, vec_sub, vec_neg,
-                    xor)
+from common import (ERR_INPUT, format_dst, gen_rand, to_le_bytes, vec_add,
+                    vec_sub, vec_neg, xor)
 from field import Field128, Field2
 import hashlib
 from xof import XofFixedKeyAes128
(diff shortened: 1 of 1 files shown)


poc/vidpf.py
```diff
@@ -116,42 +121,46 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
...
+
+        # Compute the path verifier.
+        sha3 = hashlib.sha3_256()
+        for prefix in prefixes:
+            for current_level in range(level):
+                node = prefix >> (level - current_level)
+                y =  prefix_tree_share[(node,        current_level)  ][2]
+                y0 = prefix_tree_share[(node<<1,     current_level+1)][2]
+                y1 = prefix_tree_share[((node<<1)|1, current_level+1)][2]
+                sha3.update(cls.Field.encode_vec(vec_sub(y, vec_add(y0, y1))))
```
cjpatton, line 151: @jimouris I think the math actually works out so that we don't need to conditionally invert here, as suggested in https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/20#discussion_r1362349956

poc/vidpf.py
```diff
@@ -178,7 +187,7 @@ def eval_next(cls, prev_seed, prev_ctrl, correction_word, cs_proof, level, x_les
 
         sha3 = hashlib.sha3_256()
         # pi' = H(x^{<= i} || s^i)
-        sha3.update(str(x_less_than_i).encode('ascii') + next_seed)
+        sha3.update(str(node).encode('ascii') + next_seed)
```
cjpatton, line 190: @jimouris I think this may be ambiguous: the node of the prefix tree is determined by `(node, current_level)`, not just `node`. Otherwise there are collisions.

poc/vidpf.py
jimouris: Hmm, I don't see why it checks out without negating one part: 

$y_0^γ + y_1^γ = (y_0^{γ||0} + y_0^{γ||1}) + (y_1^{γ||0} + y_1^{γ||1}) \iff y_0^γ - y_0^{γ||0} - y_0^{γ||1} = - y_1^γ + y_1^{γ||0} + y_1^{γ||1}$

poc/vidpf.py
jimouris: Ambiguous to put the node you mean? What are you suggesting here instead?

jimouris (commented):
Looks good, need to double-check the conditional negate.

poc/vidpf.py
cjpatton: I'm suggesting `node || current_level`.

poc/vidpf.py
cjpatton: Yeah I see what you mean. But this is consistent with how IDPF works in the Poplar paper: In order to make the output shares additive, you actually need to negate one of them, as we do here.

jimouris: But here we haven't negated, right? Or am I missing it?

cjpatton: I haven't added a negation in this PR

cjpatton: Sorry, II think this is what you're asking for: https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/main/poc/vidpf.py#L149

jimouris: Oh nice, yup, that's the one, so we're negating but in the VIDPF instead of here. Looks good!

jimouris (approved):

jimouris (approved):

PR #25, opened 2023-10-19
Clarify and speed up mastic implementation

hannahdaviscrypto:
Clarifying certain comments and error messages in the Mastic implementation. The only significant change is to VIDPF.eval, which now uses the INCREMENTAL_MODE flag to determine whether to establish path-verifiability for the whole tree or a single layer. This should eliminate redundant hashing in incremental mode and slightly improve efficiency.

### DIFF
--- a/poc/vdaf_mastic.py
+++ b/poc/vdaf_mastic.py
@@ -96,6 +96,11 @@ class Mastic(Vdaf):
             # One solution is to make the first level a parameter of the VDAF.
             # This is probably a good idea anyway, since it's a trade-off the
             # Aggregators will probably want to agree on anyway.
+            #
+            # Another solution is to change the formatting of agg_param to
+            # support tuples of (level, prefixes) pairs. This would allow the
+            # VDAF to validate levels '0' and '7' in a single round trip,
+            # thus supporting faststart without changing the validity condition.
             return (
                 (len(previous_agg_params) == 0 and level == 0) or \
                 previous_agg_params[0][0] == 0
@@ -158,6 +163,8 @@ class Mastic(Vdaf):
                                                         cls.ROOT_PROOF,
                                                         nonce)
             meas_share = vec_add(out_share[0], out_share[1])
+            # As a consequence of VIDPF path verifiability,
+            # meas_share contains the value of the root node in the VIDPF tree
 
             query_rand = cls.Xof.expand_into_vec(
                 cls.Flp.Field,
@@ -188,7 +195,8 @@ class Mastic(Vdaf):
         (level_proof_0, verifier_share_0) = prep_shares[0]
         (level_proof_1, verifier_share_1) = prep_shares[1]
         if level_proof_0 != level_proof_1:
-            raise Exception('output vector is not one-hot')
+            raise Exception('output vector is not one-hot or VIDPF output is \
+                            inconsistent between levels')
 
         # Finish verifying the FLP, if applicable.
         if cls.do_range_check(agg_param):
--- a/poc/vidpf.py
+++ b/poc/vidpf.py
@@ -113,8 +113,8 @@ class Vidpf:
             if prefix >= 2 ** (level+1):
                 raise ValueError("prefix too long")
 
-            # The Aggregator's output share is the value of a node of
-            # the IDPF tree at the given `level`. The node's value is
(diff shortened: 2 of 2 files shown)


poc/vdaf_mastic.py
```diff
@@ -96,6 +96,11 @@ def is_valid(cls, agg_param, previous_agg_params):
             # One solution is to make the first level a parameter of the VDAF.
             # This is probably a good idea anyway, since it's a trade-off the
             # Aggregators will probably want to agree on anyway.
+            #
+            # Another solution is to change the formatting of agg_param to
+            # support tuples of (level, prefixes) pairs. This would allow the
+            # VDAF to validate levels '0' and '7' in a single round trip,
+            # thus supporting faststart without changing the validity condition.
```
cjpatton, line 103:
```suggestion
            # thus supporting fast-start without changing the validity condition.
```

poc/vdaf_mastic.py
```diff
@@ -96,6 +96,11 @@ def is_valid(cls, agg_param, previous_agg_params):
             # One solution is to make the first level a parameter of the VDAF.
             # This is probably a good idea anyway, since it's a trade-off the
             # Aggregators will probably want to agree on anyway.
+            #
+            # Another solution is to change the formatting of agg_param to
```
cjpatton, line 100: To be honest I think yours is a much better solution than mine. Feel free to take over the `TODO`.

poc/vdaf_mastic.py
```diff
@@ -158,6 +163,8 @@ def prep_init(cls, verify_key, agg_id, agg_param,
                                                         cls.ROOT_PROOF,
                                                         nonce)
             meas_share = vec_add(out_share[0], out_share[1])
+            # As a consequence of VIDPF path verifiability,
+            # meas_share contains the value of the root node in the VIDPF tree
```
cjpatton, line 167: * Generally block comments precede the code they're commenting on.
* I think we could merge this into the block comment above `eval()`.

poc/vdaf_mastic.py
```diff
@@ -188,7 +195,8 @@ def prep_shares_to_prep(cls, agg_param, prep_shares):
         (level_proof_0, verifier_share_0) = prep_shares[0]
         (level_proof_1, verifier_share_1) = prep_shares[1]
         if level_proof_0 != level_proof_1:
-            raise Exception('output vector is not one-hot')
+            raise Exception('output vector is not one-hot or VIDPF output is \
+                            inconsistent between levels')
```
cjpatton, line 199: Otherwise this string contains a bunch of empty space
```suggestion
            raise Exception('output vector is not one-hot or VIDPF output is ' \
                            'inconsistent between levels')
```

poc/vidpf.py
```diff
@@ -113,8 +113,8 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
             if prefix >= 2 ** (level+1):
                 raise ValueError("prefix too long")
 
-            # The Aggregator's output share is the value of a node of
-            # the IDPF tree at the given `level`. The node's value is
+            # The Aggregator's output share is the value of one or more nodes of
+            # the IDPF tree at the given `level`. These nodes' values are
```
cjpatton, line 117: Break lines at 80 characters

poc/vidpf.py
```diff
@@ -142,7 +142,10 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
         # Compute the path verifier.
```
cjpatton, line 142: Please expand on this comment here to explain what portion of the prefix tree we have path verifiability.

cjpatton (commented):
Good idea, just a few minor comments.

cjpatton:
Actually, one substantive comment: if we're not verifying the path of the entire prefix tree, then we probably need to restrict the aggregation parameter in incremental mode such that the level monotonically increases.

jimouris (approved):
I agree with Chris' comments, the rest look goog to me.

hannahdaviscrypto:
> Actually, one substantive comment: if we're not verifying the path of the entire prefix tree, then we probably need to restrict the aggregation parameter in incremental mode such that the level monotonically increases.

Issue #24, opened 2023-10-19
mastic: Change agg_param format to support fast-start and/or simultaneous level computation?


hannahedavis:
In a comment, @cjpatton raised the idea of a "fast-start" mode that allows aggregators to begin aggregating at an arbitrary level with an arbitrary set of starting prefixes. Currently Mastic does not support this feature because the FLP must be verified at level zero before any further output can be validated. Since aggregators do not necessarily store information about previous rounds except when validating aggregation parameters, one suggestion was to change the aggregation parameter to include information about which level was first evaluated. The downside of this is that the aggregation parameter format would become inconsistent with Poplar, and thus introduce changes at the DAP level. 

Another suggestion, with the same downside, is to change the aggregation parameter format to a tuple of the current aggregation parameter types. Mastic could then support simultaneous aggregation of multiple different levels on the same reports. The validity condition on the new aggregation parameters would be the existing condition applied to the parameters' union. 

The one solution that doesn't require aggregation parameters to change is to fix the starting level as a parameter of the VDAF scheme.

hannahdaviscrypto:
Fast-start added to Mastic. The aggregation parameter now includes a boolean flag indicating whether the FLP should be verified, and a sequence of aggregation parameters ins considered only if one of them includes an FLP verification.

PR #27, opened 2023-10-20
 Simplify `Mastic` and add more tests

### LINKED ISSUE #19: vidpf: Can we reduce the size of the public share in "non-incremental" mode?
As of #18, `Vidpf` has a parameter called `INCREMENTAL_MODE` indicating whether it'll be used as a VIDPF or a VDPF. This is useful for `Mastic` because it allow us to decide when to check the FLP during preparation (at level `0` in incremental mode and at level `BITS - 1` otherwise).

Why plumb this down to `Vidpf`? The hope is that we can reduce the size of the public share in non-incremental mode because we don't have to distribute correction words for levels we don't evaluate. Is this true?

If so, extend `Vidpf` with this feature. Otherwise, consider hoisting `INCREMENTAL_MODE` up to `Mastic`.

cjpatton:
Closes #19.
Partially addresses #24.

Simplify the spec in two ways:

* During `Mastic.prep_init()`, call `Vidpf.eval()` exactly once. In
  order to compute the share of `beta`, modify `Vidpf` to return the sum
  of the children of the root. This is less redundant because we end up
  visiting these nodes when evaluating the prefix tree. (In effect, this
  some logic in `Mastic` into `Vidpf`.)

* Remove the `INCREMENTAL_MODE` flag and add a `do_range_check` flag to
  the aggregation parameter. If set, the Aggregators will verify the
  FLP. Accordingly, modify `Mastic.is_valid()` to enforce that the FLP
  is verified the first time (and only the first time) the reports are
  aggregated. Also, check that the level at which the reports are
  aggregated is non-decreasing.

In addition, add examples that illustrate the two main "modes" for
`Mastic`: "heavy hitters" and "labels".


poc/vidpf.py
```diff
@@ -155,7 +155,14 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
...
             (_seed, _ctrl, y, _pi_proof) = prefix_tree_share[(prefix, level)]
             out_share.append(y if agg_id == 0 else vec_neg(y))
-        return (out_share, pi_proof + path_proof)
+
+        # Compute the Aggregator's share of `beta`.
+        y0 = prefix_tree_share[(0, 0)][2]
+        y1 = prefix_tree_share[(1, 0)][2]
+        beta_share = vec_add(y0, y1)
+        if agg_id == 1:
+            beta_share = vec_neg(beta_share)
```
cjpatton, line 164: Just to follow up on the issue from the call: We need to negate aggregator 1's share; I was negating 0's.

poc/vidpf.py
```diff
@@ -155,7 +155,14 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
...
+        return (beta_share, out_share, pi_proof + path_proof)
```
jimouris, line 165: Ah, great! Just an observation/Question, in case the prefix == '', then `beta_share == out_share`, right?

In this case, would it make sense to check in mastic if level == 0 and then add the out_share[0] (left) and out_share[1] (right) to compute the beta_share? In all other cases, we don't do anything.

jimouris (changes requested):

poc/vidpf.py
jimouris: Yup, saw that :)

poc/vidpf.py
cjpatton: `level == 0` is **not** the root, but the children of the root.

jimouris: Sure, but the comment is still valid. I meant that in the case that the following is true:

```python
if cls.do_range_check(agg_param):
```

then, beta_share == out_share[0] + out_share[1]. Or not? If yes, we can skip the beta_share altogether.

cjpatton: Yes, if `prefix == [0, 1]` and `level == 0` then computing `beta_share` here is redundant and we could instead compute it the way you describe. I've updated `Vidpf.eval()` accordingly.

jimouris: Thanks, can we move it out of vidpf to mastic_vdaf? This way, if we put the check there we won't have to return both beta_share and out_share, only one would be sufficient.

cjpatton: But then we'd have to modify `Vidpf.eval()` to conditionally compute `beta_share`, right? That seems overly complicated.

jimouris: `Vidpf.eval()` can be as we had it before and we can do the negation in mastic_vdaf. We can merge this one and I can take a stab at it later.

cjpatton: Hmmm, I'm not sure exactly what you're after. Maybe tack on a commit to this branch?

jimouris: Take a look here: https://github.com/jimouris/draft-mouris-cfrg-mastic/commit/f5088c1b9292d602498cac8f12db3ee83cd8567e

I think that's sufficient, right?

cjpatton: No because `out_share[0]` is not guaranteed to be the output share for `prefix == 0`. (Similarly for `out_share[1]`.)

cjpatton: Suppose `prep_init()` is called on `prefixes == [0b11], level == 1`.

jimouris: But then `if cls.do_range_check(agg_param):` this would be false so we don't really care.

cjpatton: Take a look again: What values of `prefixes` and `level` are permitted if `INCREMENTAL_MODE==False`?

jimouris: Ah, no it wouldn't be because we have `level == cls.Vidpf.BITS-1`. But do we need that? Why not check just for the level there?

cjpatton: What do you mean check for the level?

jimouris: I meant in `do_range_check` to only check if `level == 0`. I don't see why we need the `INCREMENTAL_MODE` there or even the `level == cls.Vidpf.BITS-1` check.

Maybe it's a good idea to add some tests at some point so if changes like my proposed one don't make sense the tests should catch this.

cjpatton: Agreed, I don't think I've been clear enough about how I'd like to use Mastic beyond heavy hitters. Let me tack on some commits that will hopefully clarify.

@hannahdaviscrypto, I'm going to remove INCREMENTAL_MODE for now, which means you'll need to rework #25 a bit.

jimouris (approved):

poc/vidpf.py
jimouris: Thanks, you explained it before but I didn't fully get it. Maybe it would make sense to add it back once we have these examples/tests.

jimouris (approved):

hannahdaviscrypto (commented):
This looks good! I have some concerns that the agg_param validity check is too broad but I'm going to attempt a fix in a separate PR, possibly in concert with the tuple idea if I have enough time.

PR #28, opened 2023-10-22
poc: Harmonize naming across shard and prep

cjpatton:
Prefix variables having to do with FLP with "flp_" (likewise for VIDPF).


jimouris (approved):
Neat, thanks for doing this!

hannahdaviscrypto (approved):
I think that will clarify things a lot. Thank you!

PR #29, opened 2023-10-22
vidpf: Align API with draft

cjpatton:


poc/vidpf.py
```diff
@@ -102,6 +110,7 @@ def eval(cls, agg_id, correction_words, init_seed, level, prefixes, cs_proofs, p
         #
         # Implementation note: We can save computation by storing
         # `prefix_tree_share` across `eval()` calls for the same report.
+        pi_proof = cls.ROOT_PROOF
```
jimouris, line 113: The reason that I was passing the proof as an argument here is that we can take the previous proof and update it. Here we're only reading the ROOT_PROOF but we're not updating anywhere, right? Or I missed it?

jimouris (changes requested):

poc/vidpf.py
hannahdaviscrypto: I believe so; the ROOT_PROOF is just a constant taking the place of the empty string in the original PLASMA design.

poc/vidpf.py
hannahdaviscrypto, line 1: In main() lines 278-281 and 323-326, I don't think we need to initialize proofs to contain two copies of ROOT_PROOF. Since the class now computes that for us and initializes pi_proof in eval automatically, I think we could just set proofs = [None, None]. But it's not going to hurt anything.

hannahdaviscrypto (approved):
Looks good to me!

poc/vidpf.py
cjpatton: Good catch.

poc/vidpf.py
cjpatton: That's right, but for Mastic we weren't updating the proof either, so I expect this to be a non-breaking change.

If we think that starting evaluation with an arbitrary proof would be useful, then we can add this back to the API when we need it.

Issue #12, opened 2023-09-01
Heavy Hitters with different Thresholds


jimouris:
We can use different heavy hitter thresholds $T_i$ based on some pre-agreed strings `agg_param ∈ agg_parameters` by the servers.

Example from PLASMA paper: This can be beneficial for traffic avoidance since different roads may have different traffic densities. For example, highways are busier than smaller suburban roads. The servers can take that into consideration during evaluation, and use higher $T$ s for highways (since there are more vehicles), and lower thresholds for smaller roads.

The aggregation parameters can be tuples instead of single values:
```
aggregation_params = [(0b0000, 7), (0b0001, 6), (0b0000000, _), (0b0000001, _)]
```
meaning that the threshold $T$ for `aggregation_param = 0b0000` is 7, the $T$ for `0b0001` is 6, and so on. Nice to have: if an aggregation parameter does not have an explicit threshold (like `(0b0000000, _)`) it should inherit from its parent (so it'd be 7 in this case).

PR #38, opened 2023-10-25
Add different thresholds example. (#12)

jimouris:
Addresses https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/12

We had discussed this one before, it's not specific to Mastic but I think it's worth calling out. Right now, it has a lot of duplicate code with `example_weighted_heavy_hitters_mode`, I could merge some of the heavy hitter logic in functions and call these instead.

### DIFF
--- a/draft-mouris-cfrg-mastic.md
+++ b/draft-mouris-cfrg-mastic.md
@@ -469,17 +469,17 @@ Setting a low threshold may result in an overwhelmingly big set of heavy hitters
 starting with "000", while setting a high threshold might prune anything
 starting with "111". Consider the following examples:
 
-1. **Popular URLs:** [google.com](https://google.com/) receives a massive amount
-   of traffic whereas [duckduckgo.com](https://duckduckgo.com/) may have lower
-   traffic. To identify heavy-hitting search queries on Google, the Aggregators
-   should set a high threshold, while in queries in different search engines may
-   require lower thresholds to be considered popular.
+1. Popular URLs: `a.example.com` receives a massive amount of traffic whereas
+   `b.example.com` may have lower traffic. To identify heavy-hitting search
+   queries on `a.example.com`, the Aggregators should set a high threshold, while
+   in queries in different search engines may require lower thresholds to be
+   considered popular.
 
-1. **E-commerce:** Grocery items are essential and have a high volume of sales.
-   In contrast, electronics, though popular, usually come with a higher price
+2. E-commerce: Grocery items are essential and have a high volume of sales. In
+   contrast, electronics, though popular, usually come with a higher price
    compared to groceries. Meanwhile, luxury items command significantly higher
    prices but generally experience lower sales volumes. To identify
-   heavy-hitting grocery items on an e-commerce website, Aggregators should use
+   heavy-hitting grocery items on an e-commerce website, Aggregators could use
    different threshold for each of these categories. These thresholds are set to
    ensure that only the top-selling grocery items qualify as heavy hitters while
    electronics and luxury items are also considered heavy hitters on their own
--- a/draft-mouris-cfrg-mastic.md
+++ b/draft-mouris-cfrg-mastic.md
@@ -463,19 +463,35 @@ minimum weight rather than a minimum count. In addition:
 
 So far, we have assumed that there is a single threshold for determining which
 prefixes are "heavy". However, we can easily extend this to have different
-thresholds for different prefixes. Consider a use-case where prefixes starting
-with "000" are significantly more popular than prefixes starting with "111".
+thresholds for different prefixes. There exist use-cases where prefixes starting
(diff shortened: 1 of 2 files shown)


cjpatton (commented):
Let's add a more concrete example to motivate this. Right now it's kind of abstract (basically, we might know in advance that some prefixes are more popular than others) and I don't quite understand what purpose is served by setting different thresholds for different prefixes.

That said the example about highways versus small roads could be interesting. Perhaps elaborate on this a bit more?

Perhaps another fruitful direction: Take the "popular URLs" use case, which I think most people are familiar with. You might know that google.com is likely to be a heavy hitter; so how do you set the threshold for this?

pratiksarkar93:
Moreover, the sorting-based approaches cannot support multiple thresholds at different layers. Only the DPF-based ones (like Poplar, Doplar, and PLASMA) support it. In fact, it supports different thresholds for different prefixes (i.e., even multiple thresholds for different strings as the same level).

This technique can be useful to compute heavy-hitting selling items on Amazon based on customer's expenditures. Run the multiple-thresholds protocol over customers spending $ on Amazon purchases. The amount that needs to be spent on specific categories to consider a category as a heavy-hitter might be different. For example, grocery items will have a lower threshold than luxury items.

pratiksarkar93 (approved):

poc/vdaf_mastic.py
```diff
@@ -418,24 +416,95 @@ def example_weighted_heavy_hitters_mode():
...
-        agg_result = mastic.unshard(agg_param, agg_shares, len(measurements))
+        agg_result = mastic.unshard(agg_param, agg_shares, len(reports))
         prev_agg_params.append(agg_param)
 
         if level < bits - 1:
             # Compute the next set of candidate prefixes.
             next_prefixes = []
             for (prefix, count) in zip(prefixes, agg_result):
+                bin_prefix = '0b' + format(prefix, f'0{level}b')
+                threshold = get_threshold(thresholds, bin_prefix)
```
cjpatton, line 427: Code design: `bin_prefix` seems to represent a node in the IDPF tree. Elsewhere we represent this as the pair `(prefix, level)`. Is there a reason why the new representation is useful? I think it would be better to be consistent unless there's a reason not to.

poc/vdaf_mastic.py
```diff
@@ -352,40 +352,38 @@ def test_valid_agg_params():
...
         nonce = gen_rand(16)
         rand = gen_rand(mastic.RAND_SIZE)
         (public_share, input_shares) = mastic.shard(measurement, nonce, rand)
         reports.append((nonce, public_share, input_shares))
+    return reports
+
+
+def get_threshold(thresholds, prefix):
+    '''
+    Return the threshold of the given prefix if exists. If not, check if any of
```
cjpatton, line 369:
```suggestion
    Return the threshold of the given prefix if it exists. If not, check if any of
```

poc/vdaf_mastic.py
```diff
@@ -352,40 +352,38 @@ def test_valid_agg_params():
...
+    its prefixes exists. If not, return the default threshold.
```
cjpatton, line 370:
```suggestion
    its prefixes exist. If not, return the default threshold.
```

poc/vdaf_mastic.py
jimouris: we represent the alpha values as ints (e.g., 0b001). But here we want a dict and values such as 0b00 and 0b0 are both zero. So I'm converting these to strings "0b00" and "0b0", respectively.

draft-mouris-cfrg-mastic.md
```diff
@@ -455,6 +455,44 @@ minimum weight rather than a minimum count. In addition:
...
+
+So far, we have assumed that there is a single threshold for determining which
+prefixes are "heavy". However, we can easily extend this to have different
+thresholds for different prefixes. There exist use-cases where prefixes starting
+with "000" may be significantly more popular than prefixes starting with "111".
+Setting a low threshold may result in an overwhelmingly big set of heavy hitters
+starting with "000", while setting a high threshold might prune anything
+starting with "111". Consider the following examples:
+
+1. **Popular URLs:** [google.com](https://google.com/) receives a massive amount
```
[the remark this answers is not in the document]
cjpatton: Replace it with example.com

poc/vdaf_mastic.py
cjpatton: My opinion is that we should represent the keys in this dictionary as pairs `(prefix, level)` rather than strings. Better to have the reader need just one mental model for thinking about how to represent this value.

jimouris: Okay, done.

cjpatton (approved):
👍

jimouris:
Close https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/12

PR #40, opened 2023-10-30
Add Prio3-style joint randomness derivation

hannahdaviscrypto:


hannahdaviscrypto:
I wasn't able to run the tests yet as I'm still waiting on IT support for permission to install Sage. I'll update when that comes through.

hannahdaviscrypto:
Finally, my workspace is up and running and the tests pass.

jimouris:
I'll try to take a look over the weekend, it'll help me understand how the joint randomness is generated.

jimouris:
@hannahdaviscrypto Could you also add some text in the draft to describe the joint randomness and in which cases we will need it and in which not?

poc/vdaf_mastic.py
```diff
@@ -71,10 +88,67 @@ def shard(cls, measurement, nonce, rand):
...
+        flp_leader_proof_share = vec_sub(
+            flp_proof,
+            cls.helper_proof_share(flp_helper_proof_share_seed),
+        )
+
+        input_shares = [
+            (vidpf_init_seed[0], 
+             flp_leader_proof_share,
+             flp_leader_blind
+            ),
```
cjpatton, line 141: I think something like this would be more Python-ish
```suggestion
            (
                vidpf_init_seed[0], 
                flp_leader_proof_share,
                flp_leader_blind,
            ),
```

poc/vdaf_mastic.py
```diff
@@ -125,15 +202,30 @@ def prep_init(cls, verify_key, agg_id, agg_param,
                 cls.Flp.QUERY_RAND_LEN,
             )
 
-            flp_verifier_share = cls.Flp.query(beta_share,
+            corrected_joint_rand_seed, joint_rand_part = None, None
+            joint_rand = b''
```
cjpatton, line 206:
```suggestion
            joint_rand = []
```

poc/vdaf_mastic.py
```diff
@@ -249,9 +392,9 @@ def test_vec_encode_prep_msg(Vdaf, prep_message):
 
     @classmethod
     def with_params(cls, bits, validity_circuit):
-        if validity_circuit.JOINT_RAND_LEN > 0:
+        # if validity_circuit.JOINT_RAND_LEN > 0:
             # TODO(cjpatton) Add support for FLPs with joint randomness.
-            raise NotImplementedError()
+        #    raise NotImplementedError()
```
cjpatton, line 397: Why not just delete this instead of commenting it out?

poc/vdaf_mastic.py
```diff
@@ -600,11 +743,11 @@ def h(label):
...
-    assert agg_result == [4, 0, 3]
+    assert agg_result == [6, 0, 4]
 
 
 if __name__ == '__main__':
-    from flp_generic import Count
+    from flp_generic import Count, Sum
     from common import from_be_bytes
 
     example_weighted_heavy_hitters_mode()
```
cjpatton, line 753: In the test_vdaf() examples below, maybe add a test case with `Sum`.

cjpatton (approved):
Some nit-picky comments but looks great overall.

poc/vdaf_mastic.py
hannahdaviscrypto: I did change the example_aggregation_by_labels_mode() test case to use Sum instead of Count and changed the values a bit to fall in a wider range. Should I change that back and add a new test case just for Sum?

jimouris: I think it makes sense to use `Sum` in `example_aggregation_by_labels_mode` and have only that. Using `Count` there is trivial, I don't see any application in which we'd want to do that realistically since it can be achieved with a simple encoding more effectively (in fact we don't even need the tuple if we do `Count`).

cjpatton: For test coverage purposes we should have unit tests for Mastic with and without joint randomness. Yes it's technically redundant, given that the examples already exercise the former, but the examples aren't intended to be unit tests.

cjpatton:
Hey @hannahdaviscrypto sorry about this but I merged #41, which caused conflicts that you'll need to rebase on top of. You'll also need to address any linter warnings that CI throws at you. Take a look at .github/workflows/lint-python.yml for details. Note that `autopep8` and `isort` are tools you can install and run locally to automatically fix the warnings.

poc/vdaf_mastic.py
hannahdaviscrypto: Should be done now.

jimouris (approved):

PR #42, opened 2023-11-09
Run tests from CI

cjpatton:


jimouris (approved):
Thanks for adding this!

PR #43, opened 2023-11-10
Clean up

cjpatton:
Stacked on #40 (merge that first).

PR #44, opened 2023-11-10
Pass measurement count to `Flp.decode()`

### LINKED ISSUE #35: Full FLP support: Pass `num_measurements` to `Flp.decode()`
Some FLPs require knowing the number of measurements in order to [decode the aggregate result](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf#section-7.1.1-6.1), e.g., when computing the arithmetic mean.

[We don't support this at the moment](https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/077fb76db8a0f62f8391cb44c27c89dce05fd7ca/poc/vdaf_mastic.py#L189-L199) because we don't count how many inputs match each candidate prefix.

One way to address this to always aggregate prefix counts [the way we do for plain heavy-hitters](https://datatracker.ietf.org/doc/html/draft-mouris-cfrg-mastic#section-5.3-7).

cjpatton:
Closes #35.
Stacked on #43 (merge that first).

At the moment we simply pass `meas_count == 0` to `Flp.decode()` because we don't know the actual value of `meas_count`. This makes `Mastic` incompatible with any FLP that requires the measurement count at decoding time.

To address this, plumb a counter for each measurement from VIDPF evaluation to FLP unsharding. The value of the counter is `1` if the prefix is on the `alpha`-path and `0` otherwise. To accomplish this, we do the simplest thing: prepend a field element to each correction word.

Along the way, fix a bug in `Mastic.aggregate()` and `Mastic.unshard()` whereby the accumulator vector was initialized to the wrong length.

### DIFF
--- a/poc/vidpf.py
+++ b/poc/vidpf.py
@@ -79,7 +79,7 @@ class Vidpf:
             ctrl[0] = correct(t_0[keep], ctrl_cw[keep], ctrl[0])  # t0'
             ctrl[1] = correct(t_1[keep], ctrl_cw[keep], ctrl[1])  # t1'
 
-            w_cw = vec_add(vec_sub(beta, w_0), w_1)
+            w_cw = vec_add(vec_sub([cls.Field(1)] + beta, w_0), w_1)
             mask = cls.Field(1) - cls.Field(2) * \
                 cls.Field(ctrl[1].as_unsigned())
             for j in range(len(w_cw)):
@@ -154,8 +154,19 @@ class Vidpf:
                 (seed, ctrl, y, pi_proof) = prefix_tree_share.get(
                     (node, current_level))
 
+        # Compute the Aggregator's share of `beta`.
+        y0 = prefix_tree_share[(0, 0)][2]
+        y1 = prefix_tree_share[(1, 0)][2]
+        beta_share = vec_add(y0, y1)[1:]  # first element is the counter
+        if agg_id == 1:
+            beta_share = vec_neg(beta_share)
+
+        # Compute the Aggregator's share of the counter.
+        counter_share = y0[0] + y1[0] + cls.Field(agg_id)
+
         # Compute the path proof.
         sha3 = hashlib.sha3_256()
+        sha3.update(cls.Field.encode_vec([counter_share]))
         for prefix in prefixes:
             for current_level in range(level):
                 node = prefix >> (level - current_level)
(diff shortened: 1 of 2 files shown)


poc/vidpf.py
```diff
@@ -154,8 +154,19 @@ def eval(cls,
...
 
+        # Compute the Aggregator's share of `beta`.
+        y0 = prefix_tree_share[(0, 0)][2]
+        y1 = prefix_tree_share[(1, 0)][2]
+        beta_share = vec_add(y0, y1)[1:]  # first element is the counter
+        if agg_id == 1:
+            beta_share = vec_neg(beta_share)
+
+        # Compute the Aggregator's share of the counter.
+        counter_share = y0[0] + y1[0] + cls.Field(agg_id)
```
cjpatton, line 165: @jimouris please check carefully that this matches PLASMA.

jimouris: You are referring to this specific line, right? `counter_share = y0[0] + y1[0] + cls.Field(agg_id)`

Which PLASMA check do you have in mind?

cjpatton: > You are referring to this specific line, right? `counter_share = y0[0] + y1[0] + cls.Field(agg_id)`

Yes.

> Which PLASMA check do you have in mind?

This is the thing where we make sure the value programmed into the VIDPF (i.e., part of $\beta$) is `1`.

jimouris: Not exactly. In PLASMA we do the following checks:
* for the root level: we check that $H(1 - y_{S0}^0 - y_{S0}^1) == H(y_{S1}^0 + y_{S1}^1)$
* for all other levels: we check that $H(y_{S0}^p - y_{S0}^{p||0} - y_{S0}^{p||1}) == H(-y_{S1}^p + y_{S1}^{p||0} + y_{S1}^{p||1})$

So the check that actually checks if beta is one is the first one for the root level.

cjpatton:
Rebased.

poc/vidpf.py
cjpatton: I believe that's exactly what I'm doing here:
```python
counter_share = y0[0] + y1[0] + cls.Field(agg_id)
```
Except that the second server's share is not negated. (For some reason, the math works out this way in the current VIDPF implementation, as we noted on an earlier PR.)

jimouris (approved):

poc/vidpf.py
hannahdaviscrypto, line 165: I think this counter_share isn't getting returned by the VIDPF or included in the out_share?

hannahdaviscrypto (approved):
One possible bug in how the counter_share is used, but otherwise looks good to me. Thanks, Chris!

poc/vidpf.py
cjpatton: the `out_share` returned at the end of this function would include a share of  $1 \|| \beta$ if one of the prefixes is on the $\alpha$ path. At this point we're just checking that the programmed value is indeed $1$.

PR #46, opened 2023-12-02
vidpf: Use the `Xof` API from the base draft for hashing

### LINKED ISSUE #17: poc: Use binary encoding for SHA-3 inputs rather than ascii

cjpatton:
Closes #17.

Replace SHA-3 with `XofShake128` (SHAKE128). This has two benefits:

1. It simplifies domain separation, which will be useful if we end up needing to model instances of this object as random oracles in security proofs.

2. It allows us to easily swap `XofShake128` with `XofTurboShake128` (TurboSHAKE128) as required in the next draft (draft-irtf-cfrg-vdaf-08).

We also replace update the base "cs proof" by using the SHA256 hash of a fixed string instead of SHA3. This is because SHA256 is more likely to be implemented in common cryptographic libraries.

### DIFF
--- a/poc/vidpf.py
+++ b/poc/vidpf.py
@@ -255,7 +255,6 @@ class Vidpf:
         xof = XofShake128(seed, b'vidpf cs proof', binder)
         return xof.next(PROOF_SIZE)
 
-
     @classmethod
     def with_params(cls, f, bits, value_len):
         class VdipfWithField(cls):
--- a/poc/vidpf.py
+++ b/poc/vidpf.py
@@ -7,9 +7,13 @@ sys.path.append('draft-irtf-cfrg-vdaf/poc')  # nopep8
 import hashlib
 
 from common import (format_dst, gen_rand, to_le_bytes, vec_add, vec_neg,
-                    vec_sub, xor)
+                    vec_sub, xor, zeros)
 from field import Field2, Field128
-from xof import XofFixedKeyAes128
+from xof import XofFixedKeyAes128, XofShake128
+
+ROOT_PI_PROOF = hashlib.sha256(b"vidpf root pi proof").digest()
+
+PROOF_SIZE = 32
 
 
 class Vidpf:
@@ -17,7 +21,6 @@ class Vidpf:
 
     # Operational parameters.
     Field = None  # set by `with_params()`
-    ROOT_PROOF = hashlib.sha3_256().digest()  # Hash of the empty string
 
     # Bit length of valid input values (i.e., the length of `alpha` in bits).
     BITS = None  # set by `with_params()`
@@ -86,16 +89,10 @@ class Vidpf:
                 w_cw[j] *= mask
 
             # Compute hashes for level i
-            sha3 = hashlib.sha3_256()
-            sha3.update(str(node).encode('ascii') +
-                        to_le_bytes(i, 2) + seed[0])
-            proof_0 = sha3.digest()
-            sha3 = hashlib.sha3_256()
-            sha3.update(str(node).encode('ascii') +
-                        to_le_bytes(i, 2) + seed[1])
-            proof_1 = sha3.digest()
-
-            cs_proofs.append(xor(proof_0, proof_1))
+            cs_proofs.append(xor(
+                next_cs_proof(i, seed[0]),
+                next_cs_proof(i, seed[1]),
+            ))
             correction_words.append((seed_cw, ctrl_cw, w_cw))
 
         return (init_seed, correction_words, cs_proofs)
@@ -121,7 +118,7 @@ class Vidpf:
         #
         # Implementation note: We can save computation by storing
         # `prefix_tree_share` across `eval()` calls for the same report.
-        pi_proof = cls.ROOT_PROOF
+        pi_proof = ROOT_PI_PROOF
         prefix_tree_share = {}
         for prefix in prefixes:
             if prefix >= 2 ** (level+1):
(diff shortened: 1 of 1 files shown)


hannahdaviscrypto:
Small question: don't we need SHA3 (or another indifferentiable hash function) for the FLPs? So in the case where SHA256 is implemented but SHA3 isn't, Mastic may not be a good candidate regardless of the choice for the cs proof.

poc/vidpf.py
```diff
@@ -86,16 +89,10 @@ def gen(cls, alpha, beta, binder, rand):
...
-                        to_le_bytes(i, 2) + seed[0])
-            proof_0 = sha3.digest()
-            sha3 = hashlib.sha3_256()
-            sha3.update(str(node).encode('ascii') +
-                        to_le_bytes(i, 2) + seed[1])
-            proof_1 = sha3.digest()
-
-            cs_proofs.append(xor(proof_0, proof_1))
+            cs_proofs.append(xor(
+                next_cs_proof(i, seed[0]),
```
hannahdaviscrypto, line 93: I think the prefix/node is getting left out here.

poc/vidpf.py
```diff
@@ -161,28 +158,26 @@ def eval(cls,
...
+                path += cls.Field.encode_vec(vec_sub(y, vec_add(y0, y1)))
 
         # Compute the Aggregator's output share.
         out_share = []
         for prefix in prefixes:
             (_seed, _ctrl, y, _pi_proof) = prefix_tree_share[(prefix, level)]
             out_share.append(y if agg_id == 0 else vec_neg(y))
 
-        return (beta_share, out_share, pi_proof + path_proof)
+        return (beta_share, out_share, eval_proof(pi_proof, counter, path))
```
hannahdaviscrypto, line 180: Combining the one-hot and sum proofs into a single hash crosses the VIDPF abstraction boundary. I think it's likely okay from a security standpoint but we'll definitely want the change reflected in the paper.

poc/vidpf.py
```diff
@@ -212,20 +207,15 @@ def eval_next(cls, prev_seed, prev_ctrl, correction_word, cs_proof,
         for i in range(len(w)):
             y.append(w[i] + w_cw[i] * mask)
 
-        sha3 = hashlib.sha3_256()
         # pi' = H(x^{<= i} || s^i)
-        sha3.update(str(node).encode('ascii') +
-                    to_le_bytes(current_level, 2) + next_seed)
-        pi_prime = sha3.digest()
+        pi_prime = next_cs_proof(current_level, next_seed)
```
hannahdaviscrypto, line 211: Again, the node is getting removed here.

hannahdaviscrypto (changes requested):
I think there are a couple of changes here with implications for the proof/paper, we should confirm that we're all on the same page about them before merging.

cjpatton:
> Small question: don't we need SHA3 (or another indifferentiable hash function) for the FLPs? So in the case where SHA256 is implemented but SHA3 isn't, Mastic may not be a good candidate regardless of the choice for the cs proof.

This change is replacing SHA3 with SHAKE128, not SHA256. SHAKE128 is Keccak-based, just like SHA3.

poc/vidpf.py
cjpatton: Good catch, I completely missed this. Modified `next_cs_proofI()` to take in the node as input. Please double check how the node is encoded.

poc/vidpf.py
cjpatton: Done.

poc/vidpf.py
cjpatton: How do you mean? (What is "sum proof" here?) The VIDPF proof is meant to ensure three things:
1. The output is one-hot (`pi_proof`)
2. The value of the counter is one (`counter`)
3. Each node in the prefix tree is the sum of its children (`path`)

hannahdaviscrypto: At least in the paper, the VIDPF only guarantees 1. and 2. as you've numbered them. 3 is a separate check that PLASMA does on the VIDPF output shares, which is why it used a separate proof. This is at least partially because the path proof requires multiple VIDPF evaluations to generate, so it doesn't quite line up with the existing syntax.

poc/vidpf.py
```diff
@@ -257,6 +247,14 @@ def convert(cls, seed, binder):
         next_seed = xof.next(XofFixedKeyAes128.SEED_SIZE)
         return (next_seed, xof.next_vec(cls.Field, 1+cls.VALUE_LEN))
 
+    @classmethod
+    def next_cs_proof(cls, node, level, seed):
+        binder = to_le_bytes(cls.BITS, 2) \
+            + to_le_bytes(node, (cls.BITS + 7) // 8) \
+            + to_le_bytes(level, 2)
```
cjpatton, line 254: @hannahdaviscrypto are you happy with this?

hannahdaviscrypto:
> > Small question: don't we need SHA3 (or another indifferentiable hash function) for the FLPs? So in the case where SHA256 is implemented but SHA3 isn't, Mastic may not be a good candidate regardless of the choice for the cs proof.
> 
> This change is replacing SHA3 with SHAKE128, not SHA256. SHAKE128 is Keccak-based, just like SHA3.

Sorry, miscommunication. The comment was in reference to the ROOT_PI_PROOF only, which this commit changes from a SHA3 digest to a SHA256 digest. The initial comment suggests this is because SHA256 is more likely to be implemented, but we're going to need a Keccak implemenation regardless.

cjpatton:
> > > Small question: don't we need SHA3 (or another indifferentiable hash function) for the FLPs? So in the case where SHA256 is implemented but SHA3 isn't, Mastic may not be a good candidate regardless of the choice for the cs proof.
> > 
> > 
> > This change is replacing SHA3 with SHAKE128, not SHA256. SHAKE128 is Keccak-based, just like SHA3.
> 
> Sorry, miscommunication. My comment was in reference to the ROOT_PI_PROOF only, which this commit changes from a SHA3 digest to a SHA256 digest. The initial PR description suggests this is because SHA256 is more likely to be implemented, but we're going to need a Keccak implemenation regardless.

That's my bad, I didn't provide enough context.

We're using SHAKE, but when we align with the latest VDAF draft (draft-irtf-cfrg-vdaf-08), we will switch to TurboSHAKE. (See https://mailarchive.ietf.org/arch/msg/cfrg/cFTfEq9zWSzgnomY4TMEeXPcoVU/ for reasoning.) SHAKE and TurboSHAKE both use Keccak (as SHA3 does) but but TurboSHAKE uses half as many rounds. Thus an implementation of permutation may not be interchangeable.

In fact, the function we use to derive the root proof doesn't matter. I expect implementations to hard-code the hash in any case. We're sort of using this value as a salt: we want it to be unlikely that some other application would have the same hard-coded salt in it. We could just pick a random 32-byte number, but using a hash of a public value ensure there are no "nothing-up-my-sleeve" constants in the spec (https://words.filippo.io/dispatches/seeds-bounty/).

I'd be happy to go back to using SHA3 instead.

hannahdaviscrypto (approved):

PR #50, opened 2023-12-07
Editorial changes

divergentdave:
This makes a few editorial changes from my first read-through of the document and proof of concept implementation. See the individual commit messages for explanations.

### DIFF
--- a/draft-mouris-cfrg-mastic.md
+++ b/draft-mouris-cfrg-mastic.md
@@ -399,9 +399,8 @@ Mastic uses a combination of techniques to certify the validity of this output.
    as specified in {{vidpf-construction}} does not as specified permit
    evaluation at the root of the VIDPF tree. Instead, each Aggregator
    computes a share of `beta` by evaluating the VIDPF tree at prefixes `0`
-   and `1` and `level == 0` and adding them up. One-hot Verifiability and
-   Path Verifiability imply that the sum is equal to the Aggregator's share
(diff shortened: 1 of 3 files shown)


cjpatton (approved):

draft-mouris-cfrg-mastic.md
```diff
@@ -399,9 +399,8 @@ Mastic uses a combination of techniques to certify the validity of this output.
    as specified in {{vidpf-construction}} does not as specified permit
    evaluation at the root of the VIDPF tree. Instead, each Aggregator
    computes a share of `beta` by evaluating the VIDPF tree at prefixes `0`
-   and `1` and `level == 0` and adding them up. One-hot Verifiability and
-   Path Verifiability imply that the sum is equal to the Aggregator's share
-   of `beta`.
+   and `1` and `level == 0` and adding them up. One-hot Verifiability implies
+   that the sum is equal to the Aggregator's share of `beta`.
```
jimouris, line 403: This one is a bit tricky as one-hotness alone implies only that the sum is non-zero at one position. We need both properties together to be sure that this non-zero value is indeed `beta` (which we have checked at the root with the FLP).

jimouris (changes requested):
Looks great, thanks for reading through it! I only have the small comment about one-hotness + path verifiability.

jimouris (approved):

PR #65, opened 2024-09-12
Improve readability and add some unittests

cjpatton:

PR #66, opened 2024-09-13
Harmonize proof computation with the rest of the code

cjpatton:
Stacked on #65 (merge that first).

What we previously referred to as `pi_prime` is now the "node proof". The node proof is corrected much like the seed, control bit, and payload.

Rename `pi_proof_adjustment()` to `adjusted_proof()`, getting rid of the last reference to "pi".

Finally, replace the zero seed in `eval_proof()` and `adjusted_proof()` with the proof we want to "hash forward". As of draft-irtf-cfrg-vdaf-12, XofTurboShake128 can handle variable length seeds.

### DIFF
--- a/poc/tests/test_vidpf.py
+++ b/poc/tests/test_vidpf.py
@@ -4,10 +4,70 @@ from random import randrange
 from vdaf_poc.common import gen_rand, vec_add
 from vdaf_poc.field import Field2, Field128
 
-from vidpf import Vidpf
+from vidpf import PROOF_INIT, Vidpf
 
 
 class Test(unittest.TestCase):
+
+    def test_eval_invariants(self):
+        vidpf = Vidpf(Field128, 5, 1)
+        nonce = gen_rand(vidpf.NONCE_SIZE)
+        rand = gen_rand(vidpf.RAND_SIZE)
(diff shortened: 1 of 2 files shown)


poc/tests/test_vidpf.py
```diff
@@ -209,36 +209,53 @@ def test_malformed_correction_word_payload(self):
...
+                # NOTE Even if true, this doesn't seem to imply an attack on
+                # robustness, at least not immediately. For privacy, it's
+                # crucial to make sure the client and honest aggregator agree
+                # on the public share: if the attacker can feed the aggregator
+                # a correction word that's been mangled in this way, then it
+                # can use it to distinguish whether a given prefix is on-path.
+                if vidpf.is_prefix(prefix, alpha, level):
+                    self.assertFalse(valid)
+                # else:
+                #     self.assertTrue(valid)
```
cjpatton, line 261: @jimouris I haven't quite figured out where my analysis is going wrong.

cjpatton: I figured out what's going on :)

jimouris: Was the test wrong?

cjpatton:
Rebased.

jimouris (approved):

PR #67, opened 2024-09-13
Change traversal order, more encapsulation

### LINKED ISSUE #49: Swap loop order in computation of path proof
Currently, path proof computation iterates over prefixes in the outer loop, and levels in the inner loop. Switching this would improve the data locality of implementations, while still committing to all the same data. In particular, the first few iterations of an outer level loop would often hash the same field vector in consecutive iterations of the inner loop. (because, for example, at level 0, it would process node 0 several times, and then process node 1 several times) The field vector arithmetic could even be conditionally reused across inner loop iterations for a further speedup.

cjpatton:
Closes #49.

Plus, add more encapsulation so that it's easier to refactor.

cc/ @divergentdave

### DIFF
--- a/poc/tests/test_vidpf.py
+++ b/poc/tests/test_vidpf.py
@@ -149,6 +149,41 @@ class Test(unittest.TestCase):
             [Field128(0), Field128(0)],
         ])
 
+    def test_exhuastive(self):
+        """
+        Evaluate all possible prefixes and ensure the on-path prefixes (and
+        only those prefixes) evaluate to `beta`.
+        """
+        vidpf = Vidpf(Field128, 5, 1)
+        alpha = randrange(2**vidpf.BITS)
+        nonce = gen_rand(vidpf.NONCE_SIZE)
+        rand = gen_rand(vidpf.RAND_SIZE)
+        (pub, keys) = vidpf.gen(alpha, [Field128(13)], nonce, rand)
+
+        for level in range(vidpf.BITS):
+            prefixes = tuple(range(2**level))
+            out_shares = []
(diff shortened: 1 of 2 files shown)


poc/tests/test_vidpf.py
```diff
@@ -152,6 +149,37 @@ def test(self):
             [Field128(0), Field128(0)],
         ])
 
+    def test_exhuastive(self):
```
jimouris, line 152: Could you add a comment here on what the test does so it's easier for people to read it? It should be clear already but more comments would be good.

poc/vidpf.py
```diff
@@ -227,34 +229,29 @@ def eval(self,
...
-                for inner_idx in [idx, idx.sibling()]:
-                    # Compute the value for the node and its sibling. The
+        for i in range(level+1):
+            for prefix in prefixes:
+                if prefix not in range(2 ** (level+1)):
+                    raise ValueError("prefix too long")
+
+                # Compute the entry for `prefix`. To do so, we first need to
+                # look up the parent node.
+                idx = PrefixTreeIndex(prefix >> (level - i + 1), i - 1)
```
jimouris, line 241: I'm a bit confused with the `+ 1` and `i - 1`, could you explain it in a comment?

jimouris (approved):

poc/tests/test_vidpf.py
cjpatton: Done.

poc/vidpf.py
cjpatton: Done.

PR #69, opened 2024-09-13
Use `bool` for control bits instead of `Field2`

cjpatton:
Stacked on #68.

This makes the code a bit clearer.

Issue #34, opened 2023-10-24
Harmonize VIDPF with changes to IDPF in draft-irtf-cfrg-vdaf-12


cjpatton:
We would like these to be aligned as possible. We're still optimizing IDPF here and there, e.g., https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/303. We should decide if we're OK with these optimizations.

cjpatton:
We should expect a speed-up for Mastic here: https://github.com/divviup/libprio-rs/pull/859#issuecomment-1834297018

cjpatton:
@armfazh this one is also assigned to you. Basically we took an optimization that reducees the number of AES calls per level from 3 to 2. This improved speed quite a bit, so we want this change for VIDPF as well.

jimouris:
@cjpatton close this one?

cjpatton:
Not yet, -12 is still work in progress. @divergentdave is working on some changes to IDPF that we'll want to take in VIDPF: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/poc/vdaf_poc/vdaf_poplar1.py#L191

PR #70, opened 2024-09-13
Clean up TODOs

cjpatton:
Stacked on #69.
Partially addresses #34.

* Harmonize code used for domain separation. At the moment, we overload `Vdaf.domain_separation_tag()` to use our own `VERSION`, but we invoke `format_tag()`, which uses the version from the VDAF draft. Eventually we'll want to use our own version for both; for the moment, the minimally invasive change is to just use upstream.

* Add a NOTE about an apparent bug that's blocking us from making the XOF generic. This may be a `Vdaf` bug or a bug with `Generic` itself. While at it, deduplicate the code for deriving the prove randomness.

* Modify query randomness derivation by binding to the level. We only generate query randomness once, so this may not be strictly necessary. But this does provide domain separation whenever the aggregation starts at different levels.

    * Note: The TODO suggested binding to the aggregation parameter as
      well, but this doesn't seem necessary as the FLP can only be used
      once.

* Align correction word encoding with the IDPF specification in the base draft. Note that the considerations are a little different because we don't have a concept of "inner field" versus "leaf field", but consistency is nice.

* Rename the "counter" to the "counter check" and the "path" to the "payload check".

* Refer to the "VIDPF proof" from {{MST24}} as the "path proof". This is to distinguish it from the "evaluation proof", which includes the path proof and the counter and payload checks.

### DIFF
--- a/poc/mastic.py
+++ b/poc/mastic.py
@@ -2,7 +2,7 @@
 
 from typing import Optional, Sequence, TypeAlias, TypeVar, cast
 
-from vdaf_poc.common import (byte, concat, front, to_be_bytes, vec_add,
+from vdaf_poc.common import (byte, concat, front, to_le_bytes, vec_add,
                              vec_sub, zeros)
 from vdaf_poc.field import NttField
 from vdaf_poc.flp_bbcggi19 import FlpBBCGGI19, Valid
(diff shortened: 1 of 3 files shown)


jimouris (approved):
looks good!

PR #74, opened 2024-09-17
Clean up domain separation

### LINKED ISSUE #71: Add draft versioning
We're currently use `VERSION` from draft-irtf-cfrg-vdaf for domain separation. Eventually we'll need to control versioning here so that we can break compatibility across drafts.

cjpatton:
Closes #71.

Replace `Vdaf.domain_separation_tag()` and `format_dst()` with a method `dst()` defined here and that uses the version of our document rather than draft-irtf-cfrg-vdaf.

### DIFF
--- a/poc/dst.py
+++ b/poc/dst.py
@@ -1,0 +1,30 @@
+"""
+Domain separation tags for all XOF usages, collected all in one place in order
+to make it easier to check that they're all distinct.
+"""
+
+from vdaf_poc.common import byte, to_le_bytes
+
+# Mastic
+USAGE_PROVE_RAND: int = 0
+USAGE_PROOF_SHARE: int = 1
+USAGE_QUERY_RAND: int = 2
+USAGE_JOINT_RAND_SEED: int = 3
+USAGE_JOINT_RAND_PART: int = 4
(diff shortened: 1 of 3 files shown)


poc/dst.py
```diff
@@ -0,0 +1,30 @@
...
+USAGE_EVAL_PROOF: int = 9
+USAGE_EXTEND: int = 10
+USAGE_CONVERT: int = 11
+
+
+def dst(usage: int) -> bytes:
+    # The version of this document. This should be `0` until the document
+    # gets adopted, at which point this should be synchronized with the
+    # latest wire-breaking working group draft.
+    VERSION: int = 0
```
jimouris, line 29: I don't understand the comment above. Shouldn't this version match https://datatracker.ietf.org/doc/draft-mouris-cfrg-mastic? (i.e., `02`).

jimouris (approved):

poc/dst.py
cjpatton: If/when this gets adopted, it'll be as a new draft, which would be called `draft-irtf-cfrg-mastic-00`.

PR #75, opened 2024-09-20
Clean up Mastic a bit

cjpatton:
* Fix some awkwardness in the control flow that was revealed when we added type enforcement (296698483fc89c3bd140cbc8a833afecae9b94d8). The main goal was to reduce the number of type assertions we needed.

* Remove "flp_" and "vidpf_" prefixes from variable names wherever which primitive they're associated with is clear in context.

* Avoid nesting the term "public share". In other words, Mastic's public is composed of VIDPF correction words and FLP joint rand parts; make these components explicit. This avoids confusing variable aliases, like `flp_public_share = joint_rand_parts`.

* Make `is_valid()` stricter by preventing aggregating at the same level of the VIDPF tree more than once. We don't know if this is secure, and in any case, this behavior isn't covered by the current proof. Also, try and make the spec of `is_valid()` a little simpler.

* Align the FLP joint randomness confirmation with @hannahdaviscrypto's implementation. Namely, the prep share includes the aggregator's joint randomness part, and `prep_shares_to_prep()` combines them into the true joint randomness seed. Then `prep_next()` just confirms that the true joint randomness seed matches the one it computed during `prep_init()`.

* Add a unit test that exercises Mastic with joint randomness. We had been testing this, with the `Sum` circuit; but the `Sum` circuit no longer uses joint randomness in the latest version of the base draft.

* Improve `is_valid()` unit tests by replacing `assert`s with `self.assertTrue()`s.

### DIFF
--- a/poc/mastic.py
+++ b/poc/mastic.py
@@ -41,11 +41,9 @@ MasticPrepState: TypeAlias = tuple[
 ]
 
 MasticPrepShare: TypeAlias = tuple[
-    bytes,  # VIDPF proof
-    Optional[tuple[
-        list[F],          # FLP verifier share
-        Optional[bytes],  # FLP joint randomness part
-    ]],
+    bytes,              # VIDPF proof
+    Optional[list[F]],  # FLP verifier share
+    Optional[bytes],    # FLP joint randomness part
 ]
 
 MasticPrepMessage: TypeAlias = Optional[bytes]  # FLP joint rand seed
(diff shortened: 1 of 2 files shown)


poc/mastic.py
```diff
@@ -41,11 +41,9 @@
...
 
 MasticPrepShare: TypeAlias = tuple[
-    bytes,  # VIDPF proof
-    Optional[tuple[
-        list[F],          # FLP verifier share
-        Optional[bytes],  # FLP joint randomness part
-    ]],
+    bytes,              # VIDPF proof
+    Optional[list[F]],  # FLP verifier share
+    Optional[bytes],    # FLP joint randomness part
```
cjpatton, line 46: Removing the nested `Optional` improves control flow.

poc/tests/test_mastic.py
```diff
@@ -12,67 +12,60 @@ class TestValidAggParams(unittest.TestCase):
...
                 (2, (0b100,), True),
             ],
-        )
-
-        assert mastic.is_valid(
-            (2, (0b101,), False),
-            [
-                (2, (0b100,), True),
-            ],
-        )
```
cjpatton, line 75: We don't want to allow aggregating at the same level twice.

poc/mastic.py
```diff
@@ -139,67 +131,57 @@ def shard_with_joint_rand(
...
                  previous_agg_params: list[MasticAggParam],
                  ) -> bool:
-        (level, prefixes, do_range_check) = agg_param
-
         # Check that the range check is done exactly once.
-        first_level_range_check = \
-            (do_range_check and len(previous_agg_params) == 0) or \
-            (not do_range_check and
+        range_checked = \
+            (agg_param[2] and len(previous_agg_params) == 0) or \
```
jimouris, line 176: `agg_param[2]` makes me go back to see what is `agg_param[0]` and `agg_param[1]`.
Could we keep `(_level, _prefixes, do_range_check) = agg_param` ?

jimouris (approved):

poc/mastic.py
cjpatton: Done!

PR #89, opened 2024-10-19
Added function definitions, updated parameters

TrustworthyComputing:
- Updated function definitions
- Added missing parameters
- Added IANA identifier
- Updated Acks


draft-mouris-cfrg-mastic.md
```diff
@@ -205,11 +205,42 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
 
 {::boilerplate bcp14-tagged}
 
-This document uses the same conventions and definitions of {{Section 2 of
-!VDAF}}. The following functions are as defined therein:
-
-* `gen_rand(len: int) -> bytes`
-* TODO List all functions we use in alphabetical order.
+This document uses the same conventions and definitions of {{Sections 2, 6.1, 6.2, and 7.1 of !VDAF}}. The following functions are as defined therein:
+
```
cjpatton, line 209: Wrap all lines at 80 characters

draft-mouris-cfrg-mastic.md
```diff
@@ -205,11 +205,42 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+* `flp.query(meas: list[F], proof: list[F], query_rand: list[F], joint_rand: list[F], num_shares: int) -> list[F]` is a verifier's query algorithm. This is used to "query" the measurement and proof. The result of the query (i.e., the output of this function) is called the "verifier message". In addition to the measurement and proof, this algorithm takes as input the query randomness `query_rand` and the joint randomness `joint_rand`. The former is used only by the verifier. Here, num_shares specifies how many shares were generated.
+* `flp.truncate(meas: list[F]) -> list[F]` maps an encoded measurement (e.g., the bit-encoding of the measurement) to an aggregatable output (e.g., the singleton vector containing the measurement). The length of the input MUST be `MEAS_LEN` and the length of the output MUST be `OUTPUT_LEN`.
```
cjpatton, line 227: I don't think it's necessary to list functions for APIs we use. For example, it's sufficient to say something like "Mastic is defined in terms of the `Flp` interface specified in {{Section XXX of !VDAF}}. Likewise for `Field` here.

draft-mouris-cfrg-mastic.md
```diff
@@ -205,11 +205,42 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+* `gen_rand(len: int) -> bytes` returns an array of random bytes generated by a cryptographically secure pseudorandom number generator (CSPRNG). The length of output MUST be `len`.
+* `len(obj: Sized) -> int` returns the number of items in `obj`. The object argument can be a sequence (e.g., string, list, or tuple) or a collection (e.g., dictionary, set).
+* `list.append(elem: Any) -> None` adds a single element `elem` to the end of a `list` instance.
+* `list.copy() -> list` returns a shallow copy of a `list` instance.
+* `range(stop)` or `range(start, stop[, step])` is the range function from the Python standard library. The one-argument form returns the integers from zero (inclusive) to `stop`, exclusive. The two- and three-argument forms allow overriding the start of the range and overriding the step between successive output values.
+* `set(iterable: Optional[Iterable] = None) -> set` creates a new set object, which is an unordered collection of unique elements. The optional `iterable` argument (e.g., list, tuple, or string) is used to initialize the set. If no argument is provided, an empty set is created.
```
cjpatton, line 234: I don't think we use this in the Mastic spec, but worth double checking.

draft-mouris-cfrg-mastic.md
```diff
@@ -205,11 +205,42 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+* `XofTurboShake128(seed: bytes, dst: bytes, binder: bytes) -> Xof` is a XOF wrapper for TurboSHAKE128 that is initialized with a given seed, a domain separation tag, and binder string, and returns a `Xof` instance. The length of `seed` will typically be SEED_SIZE, but some XOFs may support multiple sizes. The `seed` MUST be generated securely (i.e., it is either the output of a CSPRNG or a previous invocation of the XOF). The length of the domain separation string `dst` passed to XofTurboShake128 MUST NOT exceed 65535 bytes.
+* `Xof.next(length: int) -> bytes` returns the next `length` bytes of output of a `Xof` instance.
+* `xor(left: bytes, right: bytes) -> bytes` returns the bitwise XOR of `left` and `right`. An exception is raised if the inputs are not the same length.
```
cjpatton, line 242: likewise here: we don't need to list the `Xof` APi.

draft-mouris-cfrg-mastic.md
```diff
@@ -270,6 +301,10 @@ MUST be in range `[0, 2^16 - 11)`.
 | `BITS: int`       | bit length of the input string `alpha`                | set by constructor                                         |
 | `VALUE_LEN: int`  | length of `beta`                                      | set by constructor                                         |
 | `field: type[F]`  | class object for the field ({{Section 6.1 of !VDAF}}) | set by constructor                                         |
+| `Field.MODULUS: int`   | prime modulus defining the field arithmetic           | to be defined                                              |
+| `MEAS_LEN: int`        | the length of the encoded measurement                 | to be defined                                              |
+| `OUTPUT_LEN: int`      | the length of the aggregatable output              | to be defined                                              |
+| `SEED_SIZE: int`       | the size (in bytes) of the XOR seed                   | 32 for `XofTurboShake128` or 16 for `XofFixedKeyAes128`
```
cjpatton, line 307: I'm not quite sure why `MEAS_LEN` and `OUTPUT_LEN` are relevant to VIDPF. Also, why add `Field.MODULUS` here?

draft-mouris-cfrg-mastic.md
```diff
@@ -1266,14 +1301,25 @@ security analysis of Mastic is provided in {{MPDST24}}.
 
 # IANA Considerations
 
-TODO
+IANA is requested to add a new identifier to the "Verifiable Distributed Aggregation Functions (VDAF)" registry, that is described in {{Section 10 of !VDAF}}. In short, VDAF identifiers are used for domain separation, which guards against failures of entropy sources, by ensuring that invocations of different VDAFs use different derived values, even if they are invoked with the same underlying random data.
+
+The new requested entry to the VDAF Identifier registry is as follows:
```
cjpatton, line 1306: nit: A little more concise of an instruction for IANA would be the following (avoids giving the registry more than one name):
```suggestion
The new requested entry to the registry is as follows:
```

draft-mouris-cfrg-mastic.md
```diff
@@ -1266,14 +1301,25 @@ security analysis of Mastic is provided in {{MPDST24}}.
...
+
+     +===============+=======================+======+===============+
+     | Value         | Scheme                | Type | Reference     |
+     +===============+=======================+======+===============+
+     | 0x00000007    | Mastic                | VDAF | Section 4     |
+     |               |                       |      | of RFC XXXX   |
```
cjpatton, line 1312: This still needs discussion, but we may not want to have a single code point for each Mastic variant. Here the "variant" is determined by the circuit that defines weight validity: `MasticCount`, `MasticSum`, etc.

draft-mouris-cfrg-mastic.md
cjpatton: Oops, forgot this is what we called it in the base draft. Disregard.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: !VDAF also lists basic functions (like range)

draft-mouris-cfrg-mastic.md
TrustworthyComputing: We are listing all the functions we use so the description is self contained (most are also in the !VDAF).

draft-mouris-cfrg-mastic.md
TrustworthyComputing: We are listing all the functions we use so the description is self contained.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: We mention MEAS_LEN on L725 and OUTPUT_LEN on L1175. Since they are mentioned in the draft text, we need to define. Where are we defining field.MODULUS?

draft-mouris-cfrg-mastic.md
cjpatton: I don't think the document needs to be completely self-contained. For example, we don't describe how `Flp` is implemented, i.e., how the ZKPs are constructed.

Moreover I think it's counter-productive if we try to make things self-contained. {{!VDAF}} is not yet complete and is subject to change, so the less we try to copy and paste from there, the less work we have to do to keep them in-sync.

draft-mouris-cfrg-mastic.md
cjpatton: My point is that we don't need to define it  unless we actually use it. Do we?

draft-mouris-cfrg-mastic.md
cjpatton: `MEAS_LEN` and `OUTPUT_LEN` are parameters of `flp`, not of `vidpf`, which is why they're not relevant here.

draft-mouris-cfrg-mastic.md
cjpatton: We should pick a range from the "reserved for private use" range.

However, I don't think we should include this table in this PR. We first need to pick some Mastic variants: see #91.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Sure we do (using 8fa0a9c):
append: L458
copy: 382
range: L385
set: L500
!VDAF also defines such functions: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/bd983637a87a7fa8353db5771c184585b2550271/draft-irtf-cfrg-vdaf.md?plain=1#L858

draft-mouris-cfrg-mastic.md
TrustworthyComputing: I think this depends on how we want to reference them; someone reading the draft should either see a pointer, or a definition. These are also defined here: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/bd983637a87a7fa8353db5771c184585b2550271/draft-irtf-cfrg-vdaf.md?plain=1#L2520

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Well, we are addressing a TODO that says _"List all functions we use in alphabetical order."_ We just need to define what is "all functions". I think every function that we explicitly use in `draft-mouris-cfrg-mastic.md` satisfies this definition (unless a function is defined elsewhere in the text, like `gen` for example). Otherwise, where do you draw the line of what is copied here vs. referenced in !VDAF. I am open to alternative views of what "all functions" should be here.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Hmm, the current guidance in !VDAF is _'The "Reserved for Private Use" code points should thus be used judiciously, because they provide no defense against such collisions'_ which may not be applicable here. Since we are proposing new VDAF variants, shouldn't we continue the numbering past `0x00000006` (i.e., just like Prio3* uses five identifies in the regular range)?

draft-mouris-cfrg-mastic.md
cjpatton: I was thinking a subset of the functions in [this list](https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-2-11). Note that the TODO says "{{Section 2 of !VDAF}}", i.e., refers to the conventions in {{!VDAF}}.

draft-mouris-cfrg-mastic.md
cjpatton: Ack, thanks for verifying.

draft-mouris-cfrg-mastic.md
cjpatton: I guess I'm not clear on what we're trying to accomplish. Currently this table lists the parameters of VIDPF; if we also want to have a table of FLP parameters, then we either need to change the caption of this table or create a new table elsewhere.

There is already a paragraph in {{conventions}} in our draft with a rather terse description of FLP syntax. Perhaps we should just expand on that?

draft-mouris-cfrg-mastic.md
cjpatton: Great question.

The point I'm trying to make is that we don't have that much control over which codepoint gets assigned to Mastic. Besides the base draft and Mastic, there are two other documents floating around describing VDAFs, and each will need a codepoint or a set of codepoints. Which document gets `0x00000007` depends largely on who gets to RFC first.

Of course, we could just pick codepoints in the IANA range and change them if they get allocated by someone else, but it's better for everyone if we can avoid the collision altogether.

For the moment, the only we reason we need codepoints is to generate test vectors. My suggestion would be to choose from the private range; as we get closer to finishing, we can ask IANA for new codepoints from the standardized range.

But again, I think it's too early to create this table.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: I guess it is a bit unclear what are the criteria for inclusion in [this list](https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-2-11). For example, it can be any function that needs to be imported from `common.py`. Or is it more subjective (i.e., _here are some functions that we often use_)?

In our case, I think it is reasonable to list the imported functions that are referenced in our text (along with basic functionalities). As a reader, I find it useful to have these definitions in one place (but personal preferences can differ on this). If it makes more sense, we can split the list to basic functions and APIs.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Alright, I think we can use `0xFFFF0001`, `0xFFFF0002` and `0xFFFF0003` as placeholders for the first 3 Mastic variants. If we settle on these variants, I don't think there is a downside to include the range table (assuming we will use these points in the test vectors).

cjpatton: Works for me.

Note: it looks like you've copy-pasted the table from the *.txt file. The table will need to be formatted properly so that `xml2rfc` formats it properly for other targets, like html.

You can copy and paste from here: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/bd983637a87a7fa8353db5771c184585b2550271/draft-irtf-cfrg-vdaf.md?plain=1#L5905-L5916

draft-mouris-cfrg-mastic.md
cjpatton: Works for me.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: ok, will wait for #91 to be resolved regarding the variants.

draft-mouris-cfrg-mastic.md
cjpatton: > I guess it is a bit unclear what are the criteria for inclusion in [this list](https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-2-11). For example, it can be any function that needs to be imported from `common.py`. Or is it more subjective (i.e., _here are some functions that we often use_)?

This is the complete list of basic functions that the specifications of Prio3 and Poplar1 use, but we don't explicitly define in full detail. What they do _should_ be clear from the description; if anything is unclear, let me know :)

> In our case, I think it is reasonable to list the imported functions that are referenced in our text (along with basic functionalities). As a reader, I find it useful to have these definitions in one place (but personal preferences can differ on this). If it makes more sense, we can split the list to basic functions and APIs.

At a minimum, we should split the basic functions and APIs into separate lists. If we're going to re-define the APIs (by this I mean the `Field`, `Xof`, and `Flp` classes defined in {{!VDAF}}), then they also need to be more complete then they are currently. Each API is consists not just of instance methods, e.g., `flp.encode()`, but potentially class methods or constructors, e.g., `XofTurboShake128()`. They also define constants, e.g., `flp.MEAS_LEN` and have generic parameters, e.g., `F`.

But again, I don't think we should be re-defining APIs. Not only does this make our jobs harder in the short term, I don't think it helps much with an RFC. These documents are not meant to be self-contained as long as dependencies have clear refernces.

draft-mouris-cfrg-mastic.md
jimouris: #91 is merged, we have five variants: https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/main/poc/mastic.py#L512

draft-mouris-cfrg-mastic.md
```diff
@@ -99,7 +100,8 @@ informative:
     target: https://ia.cr/2024/221
 
   RZCGP24:
-    title: "Private Analytics via Streaming, Sketching, and Silently Verifiable Proofs"
+    title: "Private Analytics via Streaming, Sketching, and Silently Verifiable
+     Proofs"
```
jimouris, line 104: same

draft-mouris-cfrg-mastic.md
```diff
@@ -217,10 +219,56 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
 {::boilerplate bcp14-tagged}
 
 This document uses the same conventions and definitions of {{Section 2 of
-!VDAF}}. The following functions are as defined therein:
-
-* `gen_rand(len: int) -> bytes`
-* TODO List all functions we use in alphabetical order.
+  !VDAF}}. The following basic functions are as defined therein:
+
+* `bool(val: Any) -> bool` converts a value `val` to a Boolean.
```
jimouris, line 224: Do we need to list all these (bool, byte, etc)? My impression was that we wanted just the API functions. @cjpatton WDYT?

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
 "prep share", and
 "report".
 
-Mastic uses finite fields as specified in {{Section 6.1 of !VDAF}}. We usually
-denote a finite field by `F` and its Python class object, a subclass of
-`Field`, as `field: type[F]`.
+Mastic uses finite fields of prime order as specified in {{Section 6.1 of
+  !VDAF}}. We usually denote a finite field by `F` and its Python class object,
```
jimouris, line 289: there is an odd indent here.

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+
+  | Parameter           | Description                                           | Value                   |
+  |:--------------------|:------------------------------------------------------|:------------------------|
+  | `field: type[F]`    | class object for the field ({{Section 6.1 of !VDAF}}) | set by constructor      |
+  | `Field.MODULUS: int`| prime modulus defining the field arithmetic           | to be defined           |
+  {: #field-params title="Basic Field parameters."}
+
+
+Mastic uses Extendable Output Functions (XOFs) as specified in {{Section 6.2 of
+  !VDAF}}. The following XOF API functions referenced in this document are as
```
jimouris, line 313: fix indent

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+  * `XofTurboShake128(seed: bytes, dst: bytes, binder: bytes) -> Xof` is a XOF
+    wrapper for TurboSHAKE128 that is initialized with a given seed, a domain
+    separation tag, and binder string, and returns a `Xof` instance. The length of
+    `seed` will typically be SEED_SIZE, but some XOFs may support multiple sizes.
+    The `seed` MUST be generated securely (i.e., it is either the output of a
+    CSPRNG or a previous invocation of the XOF). The length of the domain
+    separation string `dst` passed to XofTurboShake128 MUST NOT exceed 65535
+    bytes.
+  * `Xof.next(length: int) -> bytes` returns the next `length` bytes of output of
+    a `Xof` instance.
```
jimouris, line 333: I'm also in favor of not listing these as they're subject to change in VDAF so we'd also need to update accordingly. I like what you have above:
```
Mastic uses Extendable Output Functions (XOFs) as specified in {{Section 6.2 of !VDAF}}. 
```

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+The following parameters are referenced in this document:
+
+| Parameter           | Description                                           | Value                                                      |
+|:--------------------|:------------------------------------------------------|:-----------------------------------------------------------|
+| `SEED_SIZE: int`    | the size (in bytes) of the XOR seed                   | 32 for `XofTurboShake128` or 16 for `XofFixedKeyAes128`    |
+{: #xof-params title="XOF parameters."}
+
+
+Mastic uses Fully Linear Proof (FLP) Systems as specified in {{Section 7.1 of
+  !VDAF}}. The following FLP API functions referenced in this document are as
```
jimouris, line 344: remove indent

draft-mouris-cfrg-mastic.md
```diff
@@ -1300,14 +1463,14 @@ here.
...
-reporting mechanism like NEL, these events would only manifest in the server's
-telemetry as a drop in overall traffic.
+that occur while attempting to establish a connection to a server {{W3C23}}.
+Some of these errors are visible to the server, but not all: failures in DNS,
+TCP, TLS, and HTTP can occur without the server having any visibility into the
+issue. A small amount of connection errors is expected, even under normal
+operating conditions; but a sudden, substantial increase in errors may be an
+indication of an outage, or a configuration issue impacting millions of users.
+Without a reporting mechanism like NEL, these events would only manifest in the
+server's telemetry as a drop in overall traffic.
```
jimouris, line 1473: why change these?

draft-mouris-cfrg-mastic.md
```diff
@@ -1433,7 +1596,7 @@ described in {{Section 8 of !VDAF}}, except that the threshold represents a
 minimum weight rather than a minimum count. In addition:
 
 1. The Aggregators MUST perform the range check (i.e., verify the FLP) at the
-   first round of aggregation and remove any invalid reports before proceeding.
+first round of aggregation and remove any invalid reports before proceeding.
```
jimouris, line 1599: we need the indent here to keep the enumeration

draft-mouris-cfrg-mastic.md
```diff
@@ -1583,12 +1746,12 @@ the aggregation result (simply by adding to their own aggregation shares)
...
-flavor involves three Aggregators, where every pair of Aggregators communicate
-over a different channel. In essence, each pair of Aggregators will run one
-session of the VDAF with unique randomness but on the same Client measurement.
-The following changes are necessary:
+lifts the two-party semi-honest secure PLASMA to the three-party maliciously
+secure setting. Rather than having two Aggregators as in the previous setting,
+this flavor involves three Aggregators, where every pair of Aggregators
+communicate over a different channel. In essence, each pair of Aggregators will
+run one session of the VDAF with unique randomness but on the same Client
+measurement. The following changes are necessary:
```
jimouris, line 1754: why change these?

draft-mouris-cfrg-mastic.md
```diff
@@ -77,7 +77,8 @@ informative:
     target: https://iacr.org/cryptodb/data/paper.php?pubkey=31935
 
   MST24:
-    title: "PLASMA: Private, Lightweight Aggregated Statistics against Malicious Adversaries"
+    title: "PLASMA: Private, Lightweight Aggregated Statistics against Malicious
```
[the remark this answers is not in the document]
TrustworthyComputing: >80

draft-mouris-cfrg-mastic.md
TrustworthyComputing: >80

draft-mouris-cfrg-mastic.md
TrustworthyComputing: We use them in this .md. We list basic function like !VDAF does

draft-mouris-cfrg-mastic.md
TrustworthyComputing: If the API changes, so will our code that references them in this .md. With >12 iterations already, are you expecting major changes in the !VDAF API soon?

draft-mouris-cfrg-mastic.md
TrustworthyComputing: >80

draft-mouris-cfrg-mastic.md
TrustworthyComputing: >80

draft-mouris-cfrg-mastic.md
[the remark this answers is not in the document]
cjpatton: Ah, sorry I could have been more clear about this. This part of the file is YAML, so long lines shouldn't be wrapped. My suggestion was about the markdown part of the file, where it's conventional to wrap at 80 characters. (This makes it easier to review PRs.)

draft-mouris-cfrg-mastic.md
cjpatton: I agree with @jimouris. Yes, we use `bool`'s constructor, but if you understand what `bool` is, then I think think you know what `bool(x)` means since it's a primitive type in python.

draft-mouris-cfrg-mastic.md
```diff
@@ -217,10 +219,56 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
-* TODO List all functions we use in alphabetical order.
+  !VDAF}}. The following basic functions are as defined therein:
+
+* `bool(val: Any) -> bool` converts a value `val` to a Boolean.
+* `byte(number: int) -> bytes` returns the encoding of the input `int` as a
+  byte.
+* `bytearray(source: Union[str, bytes, bytearray, Iterable[int]]) -> bytearray`
+  returns a mutable sequence of bytes.
+* `bytes(source: Union[str, bytes, bytearray, Iterable[int]]) -> bytes` returns
+  a immutable sequence of bytes.
```
cjpatton, line 230: Similar to https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/89/files?diff=unified&w=0#r1823303860, I don't think we need to define all Python functions we use. It's sufficient to say "algorithms in this document are specified in Python version 3.12".

Note that we make a few exceptions to this in the base draft, where folks have specifically asked to clarify something.

Likewise below.

draft-mouris-cfrg-mastic.md
```diff
@@ -217,10 +219,56 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+* `cast(typ: Type[T], val: Any) -> T` casts a value `val` to a specific type
+  `typ`
+* `concat(parts: list[bytes]) -> bytes` returns a concatenated byte string from
+  a list of bytes.
+* `dst(usage: int) -> bytes` returns a string of bytes encoding the name of the
+  VDAF (i.e., "mastic"), the version number, and the usage type.
```
cjpatton, line 236: This is a function we define in this document. Why list it here?

Likewise below.

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
-`Field`, as `field: type[F]`.
+Mastic uses finite fields of prime order as specified in {{Section 6.1 of
+  !VDAF}}. We usually denote a finite field by `F` and its Python class object,
+  a subclass of `Field`, as `field: type[F]`. The following basic API functions
+  referenced in this document are as defined therein:
+
+* `Field(integer: int) -> Field` returns an integer represented as a field
+  element. The value of integer MUST be non-negative and less than
+  Field.MODULUS.
+* `Field.as_unsigned() -> int` returns the integer representation of a field
```
cjpatton, line 296: * This has been changed to `int()` in the upcoming draft.
* We use capitalization to distinguish between different types of Python object: `Field` indicates this is a class; `field` would indicate an instance of that class. (Though `elem` may be better variable name)

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
 
-Mastic uses finite fields as specified in {{Section 6.1 of !VDAF}}. We usually
-denote a finite field by `F` and its Python class object, a subclass of
-`Field`, as `field: type[F]`.
+Mastic uses finite fields of prime order as specified in {{Section 6.1 of
+  !VDAF}}. We usually denote a finite field by `F` and its Python class object,
+  a subclass of `Field`, as `field: type[F]`. The following basic API functions
+  referenced in this document are as defined therein:
+
+* `Field(integer: int) -> Field` returns an integer represented as a field
```
cjpatton, line 293: Why `Field` here but `Self` below?

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+* `Field.zeros(length: int) -> list[Self]` returns a vector of zeros of size
+  `length`.
+
+The following parameters are also referenced:
+
+  | Parameter           | Description                                           | Value                   |
+  |:--------------------|:------------------------------------------------------|:------------------------|
+  | `field: type[F]`    | class object for the field ({{Section 6.1 of !VDAF}}) | set by constructor      |
+  | `Field.MODULUS: int`| prime modulus defining the field arithmetic           | to be defined           |
+  {: #field-params title="Basic Field parameters."}
```
cjpatton, line 309: What's the purpose of this table? I don't see how its contents relate to one another. For example, `field` is an attribute of a validity circuit, for instance.

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+    passed to XofTurboShake128 MUST NOT exceed 65535 bytes.
+  * `XofTurboShake128(seed: bytes, dst: bytes, binder: bytes) -> Xof` is a XOF
+    wrapper for TurboSHAKE128 that is initialized with a given seed, a domain
+    separation tag, and binder string, and returns a `Xof` instance. The length of
+    `seed` will typically be SEED_SIZE, but some XOFs may support multiple sizes.
+    The `seed` MUST be generated securely (i.e., it is either the output of a
+    CSPRNG or a previous invocation of the XOF). The length of the domain
+    separation string `dst` passed to XofTurboShake128 MUST NOT exceed 65535
+    bytes.
+  * `Xof.next(length: int) -> bytes` returns the next `length` bytes of output of
```
cjpatton, line 332: `xof` since this is an intance and not a class.

cjpatton (commented):
Still working on my first pass, more comnents to come

draft-mouris-cfrg-mastic.md
TrustworthyComputing: where do you draw the line on what the reader knows? Why define range() in !VDAF then, and where do you draw the line on what is clear or not?

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Does it affect rendering? https://jimouris.github.io/draft-mouris-cfrg-mastic/nekt/draft-edits-2/draft-mouris-cfrg-mastic.html

draft-mouris-cfrg-mastic.md
TrustworthyComputing: In which line is this function defined? Are you referring to the text description of what the domain separation tag is?

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Will switch to Field

draft-mouris-cfrg-mastic.md
cjpatton: We added that one because someone asked for it.

draft-mouris-cfrg-mastic.md
cjpatton: I mean that it doesn't belong in preliminaries because it is a function that's defined in our document.

In other words, we need to list this function somewhere in the document: https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/main/poc/dst.py#L25-L30

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+    SEED_SIZE, but some XOFs may support multiple sizes. The `seed` MUST be
+    generated securely (i.e., it is either the output of a CSPRNG or a previous
+    invocation of the XOF). The length of the domain separation string `dst`
+    passed to XofTurboShake128 MUST NOT exceed 65535 bytes.
+  * `XofTurboShake128(seed: bytes, dst: bytes, binder: bytes) -> Xof` is a XOF
+    wrapper for TurboSHAKE128 that is initialized with a given seed, a domain
+    separation tag, and binder string, and returns a `Xof` instance. The length of
+    `seed` will typically be SEED_SIZE, but some XOFs may support multiple sizes.
+    The `seed` MUST be generated securely (i.e., it is either the output of a
+    CSPRNG or a previous invocation of the XOF). The length of the domain
```
cjpatton, line 329: We say the same thing for the XOF above.

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+  |:--------------------|:------------------------------------------------------|:------------------------|
+  | `field: type[F]`    | class object for the field ({{Section 6.1 of !VDAF}}) | set by constructor      |
+  | `Field.MODULUS: int`| prime modulus defining the field arithmetic           | to be defined           |
+  {: #field-params title="Basic Field parameters."}
+
+
+Mastic uses Extendable Output Functions (XOFs) as specified in {{Section 6.2 of
+  !VDAF}}. The following XOF API functions referenced in this document are as
+  defined therein:
+
```
cjpatton, line 315: There's lots of redundancy in the text below. (Also, there are some errors induced by copy-pasting.) One way we might resolve them is to define the generic `Xof` API and then say that XofFixedKeyAes128 and XofTurboShake128 are instances of that API.

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Sounds good, we can include later in the doc

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+| `SEED_SIZE: int`    | the size (in bytes) of the XOR seed                   | 32 for `XofTurboShake128` or 16 for `XofFixedKeyAes128`    |
+{: #xof-params title="XOF parameters."}
+
+
+Mastic uses Fully Linear Proof (FLP) Systems as specified in {{Section 7.1 of
+  !VDAF}}. The following FLP API functions referenced in this document are as
+  defined therein:
+
+  * `flp.decide(verifier: list[F]) -> bool` is a deterministic decision algorithm
+    of a verifier that takes as input a verifier message and outputs a Boolean
```
cjpatton, line 348: nit: Use the Python type
```suggestion
    of a verifier that takes as input a verifier message and outputs a `bool`
```

cjpatton, line 348: "algorithm of a verifier that takes as input a verifier message": does the first "verifier" refer to a party in the protocol or the verifier message? If you mean the former, then I think we need to define this; if you mean the latter, then this might grammatically problematic.

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+    indicating if the measurement from which it was generated is valid.
+  * `flp.decode(output: list[F], num_measurements: int) -> AggResult` maps a sum
+    of aggregate shares to an aggregate result.
+  * `flp.encode(measurement: Measurement) -> list[F]` encodes a raw measurement
+    as a vector of field elements. The return value MUST be of length `MEAS_LEN`.
+  * `flp.prove(meas: list[F], prove_rand: list[F], joint_rand: list[F]) ->
+    list[F]` is a prover's deterministic proof-generation algorithm. Its inputs
```
cjpatton, line 355: Who is the prover?

draft-mouris-cfrg-mastic.md
```diff
@@ -237,9 +285,98 @@ This document uses the following terms as defined in {{!VDAF}}:
...
+    singleton vector containing the measurement). The length of the input MUST
+    be `MEAS_LEN` and the length of the output MUST be `OUTPUT_LEN`.
+
+We reference the following FLP parameters in this document:
+
+  | Parameter           | Description                                           | Value               |
+  |:--------------------|:------------------------------------------------------|:--------------------|
+  | `OUTPUT_LEN: int`   | the length of the aggregatable output                 | to be defined       |
+  | `MEAS_LEN: int`     | the length of the encoded measurement                 | to be defined       |
+  | `field: type[F]`    | class object for the field ({{Section 6.1 of !VDAF}}) | set by constructor  |
```
cjpatton, line 378: This is only a subset of the FLP parameters.

Why  list `field: type[F]` here and in https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/89/files#r1823348916?

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1418,40 @@ security analysis of Mastic is provided in {{MPDST25}}.
...
+| Value          | Scheme                   | Type | Reference                 |
+|:---------------|:-------------------------|:-----|:--------------------------|
+| `0xFFFF0001`   | MasticCount              | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0002`   | MasticSum                | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0003`   | MasticSumVec             | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0004`   | MasticHistogram          | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0005`   | MasticMultihotCountVec   | VDAF | {{vdaf}} of RFC XXXX     |
+{: #codepoints title="Verifiable Distributed Aggregation Function Identifiers Registry"}
+
+
```
cjpatton, line 1440: nit: remove the second blank line
```suggestion
```

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1418,40 @@ security analysis of Mastic is provided in {{MPDST25}}.
 
 # IANA Considerations
 
-TODO
+IANA is requested to add new identifiers to the "Verifiable Distributed
+Aggregation Functions (VDAF)" registry, that is described in {{Section 10 of
+!VDAF}}. In short, VDAF identifiers are used for domain separation, which
+guards against failures of entropy sources, by ensuring that invocations of
+different VDAFs use different derived values, even if they are invoked with
+the same underlying random data.
```
cjpatton, line 1426: I don't think this is needed. (Also, note that we've updated in this text in the latest version: https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-10-10.)

```suggestion
!VDAF}}.
```

draft-mouris-cfrg-mastic.md
```diff
@@ -273,14 +411,13 @@ MUST be in range `[0, 2^16 - 11)`.
...
 
-| Parameter         | Description                                           | Value                                                      |
-|:------------------|:------------------------------------------------------|:-----------------------------------------------------------|
-| `KEY_SIZE: int`   | the size of each VIDPF key                            | `XofFixedKeyAes128.SEED_SIZE` ({{Section 6.2.2 of !VDAF}}) |
-| `NONCE_SIZE: int` | the size of the VIDPF nonce                           | `KEY_SIZE`                                                 |
-| `RAND_SIZE: int`  | the number of random bytes consumed by `gen()`        | `2 * KEY_SIZE`                                             |
-| `BITS: int`       | bit length of the input string `alpha`                | set by constructor                                         |
-| `VALUE_LEN: int`  | length of `beta`                                      | set by constructor                                         |
-| `field: type[F]`  | class object for the field ({{Section 6.1 of !VDAF}}) | set by constructor                                         |
```
cjpatton, line 283: Why remove this?

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1418,40 @@ security analysis of Mastic is provided in {{MPDST25}}.
...
+
+
+| Value          | Scheme                   | Type | Reference                 |
+|:---------------|:-------------------------|:-----|:--------------------------|
+| `0xFFFF0001`   | MasticCount              | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0002`   | MasticSum                | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0003`   | MasticSumVec             | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0004`   | MasticHistogram          | VDAF | {{vdaf}} of RFC XXXX     |
+| `0xFFFF0005`   | MasticMultihotCountVec   | VDAF | {{vdaf}} of RFC XXXX     |
+{: #codepoints title="Verifiable Distributed Aggregation Function Identifiers Registry"}
```
cjpatton, line 1438:
```suggestion
{: #codepoints title="Additional codepoints for the VDAF Identifier Registry."}
```

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1418,40 @@ security analysis of Mastic is provided in {{MPDST25}}.
...
+IANA is requested to add new identifiers to the "Verifiable Distributed
+Aggregation Functions (VDAF)" registry, that is described in {{Section 10 of
+!VDAF}}. In short, VDAF identifiers are used for domain separation, which
+guards against failures of entropy sources, by ensuring that invocations of
+different VDAFs use different derived values, even if they are invoked with
+the same underlying random data.
+
+The new requested entries to the VDAF Identifier registry are as follows:
+
+
```
cjpatton, line 1430: nit: Remove double blank line
```suggestion
```

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1418,40 @@ security analysis of Mastic is provided in {{MPDST25}}.
 
 # IANA Considerations
```
cjpatton, line 1419: Let's add a TODO to this section to track replacing the codepoints with ones we actually want to request from IANA. Something like

```
> TODO The codepoints in this section are from the private codepoint range and are used for testing purposes. We will update this table with the codepoints assigned by IANA.
```

cjpatton (changes requested):

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1418,40 @@ security analysis of Mastic is provided in {{MPDST25}}.
...
 # Acknowledgments
 {:numbered="false"}
 
-TODO
+The Network Error Logging and Attribute-Based Browser Telemetry applications
+(discussed in the next sections) were first described to the authors by the PPM
+working group at IETF. The authors would like to thank Suleman Ahmad and Simon
+Friedberger for their help in fleshing out the details. D. Mouris and N.G.
+Tsoutsos would like to acknowledge the support of the National Science
+Foundation (Award 2239334).
```
cjpatton, line 1455: Suggestion: I'd list your full names here!

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Done

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+|:--------------------|:--------------|:----------------------------|
+| `XofFixedKeyAes128` | Constructor   | {{Section 6.2 of !VDAF}}    |
+| `XofTurboShake128`  | Constructor   | {{Section 6.2 of !VDAF}}    |
+| `xof.next`          | Class Method  | {{Section 6.2 of !VDAF}}    |
+{: #XOF-functionalities title="XOF Functionalities from {{!VDAF}}."}
+
+> NOTE `xof` instances use a seed of size `SEED_SIZE` (in bytes).
+
+Each invocation of a XOF is initialized with a domain separation tag set to
+`b'mastic' + to_le_bytes(VERSION, 4) + byte(usage) + ctx`, where `ctx` is the
```
cjpatton, line 285: This has been updated:
https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/fc295a793362793eebe9a6132a2b5881fb51fd91/poc/dst.py#L30-L42

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+`[0, 2^16 - 11)`.
+
+> NOTE This range was computed by taking the maximum size of the domain
+> separation tag supported by both XofFixedKeyAes128 and XofTurboShake128 and
+> subtracting the length of the prefix.
+
+For domain separation purposes, we further define the following function:
+
+* `dst(usage: int) -> bytes` returns a string of bytes encoding the name of
+the VDAF (i.e., "mastic"), the version number, and the usage type.
```
cjpatton, line 297: Less ambigous

```suggestion
* `dst(usage: int) -> bytes` returns the domain separation tag as defined above.
```

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+
+Finally, the following common functionalities are defined as follows:
```
cjpatton, line 299: Odd sentence structure ("following ... as follows")

draft-mouris-cfrg-mastic.md
```diff
@@ -369,13 +449,11 @@ def gen(self,
...
     The VIDPF key generation algorithm.
-
     Returns the public share (i.e., the correction word for each
-    level of the tree) and two keys, one fore each aggregator.
-
+    level of the tree) and two keys, one for each aggregator.
     Implementation note: for clarity, this algorithm has not been
     written in a manner that is side-channel resistant. To avoid
-    leading `alpha` via a side-channel, implementations should avoid
+    leaking `alpha` via a side-channel, implementations should avoid
```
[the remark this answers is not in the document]
cjpatton: Actually, did you get these changes from the reference code? When did you copy the code? I merged some changes today, and there might be another one on the way (#110). It would be helpful if you could update once more, otherwise I would just revert these changes.

draft-mouris-cfrg-mastic.md
```diff
@@ -369,13 +449,11 @@ def gen(self,
         ) -> tuple[list[CorrectionWord], list[bytes]]:
```
cjpatton, line 449: Text above this point (circa 391) is now out of date: the one-hotness, path, and counter consistency checks have been moved out of VIDPF into Mastic. If you'd like to update the algorithm definitions in this PR, then it would be helpful if you could also align the prose description with the spec.

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1277,37 @@ security analysis of Mastic is provided in {{MPDST25}}.
 
 # IANA Considerations
 
-TODO
+IANA is requested to add a new identifier to the "Verifiable Distributed
```
cjpatton, line 1280:
```suggestion
IANA is requested to add new identifiers to the "Verifiable Distributed
```

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1277,37 @@ security analysis of Mastic is provided in {{MPDST25}}.
...
+Aggregation Functions (VDAF)" registry, that is described in {{Section 10 of
+!VDAF}}. The new requested entry to the VDAF Identifier registry is as follows:
```
cjpatton, line 1282:
```suggestion
!VDAF}}. The new entries to the VDAF Identifier registry are as follows:
```

draft-mouris-cfrg-mastic.md
```diff
@@ -1281,14 +1277,37 @@ security analysis of Mastic is provided in {{MPDST25}}.
...
+(RFC EDITOR: Please replace "RFC XXXX" above with the RFC number assigned to
+this document.)
 
 --- back
 
 # Acknowledgments
 {:numbered="false"}
 
-TODO
+The Network Error Logging and Attribute-Based Browser Telemetry applications
```
cjpatton, line 1305: These applications aren't described until the appendix. At a minimum we need a forward reference here, but I would just cut it.

It looks like this was copied directly from the paper. Note that the ACK section of an IETF draft doesn't necessarily serve the same purpose as in a paper. I would rather acknowledge folks who have contributed to the draft, e.g., Pratik Sarkar and David Cook.

Acknowledging a funding source is definitely acceptable, but I'm not sure Dimitris is supported by NSF anymore :)

cjpatton (commented):
I'd recommend not updating the algorithms in this PR; there is at least one more change to consider before the next draft (#110).

draft-mouris-cfrg-mastic.md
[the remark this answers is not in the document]
TrustworthyComputing: I used the most recent reference code (i.e., commit fc295a7)

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
 
-This document uses the same conventions and definitions of {{Section 2 of
+This document uses the same conventions and definitions as {{Section 2 of
 !VDAF}}. The following functions are as defined therein:
 
-* `gen_rand(len: int) -> bytes`
-* TODO List all functions we use in alphabetical order.
+| Functionality     | Type      | Definition            |
+|:------------------|:----------|:----------------------|
+| `byte`            | Function  | {{Section 2 of !VDAF}}|
```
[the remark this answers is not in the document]
TrustworthyComputing: Done

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+| `cast`            | Function  | {{Section 2 of !VDAF}}|
+| `concat`          | Function  | {{Section 2 of !VDAF}}|
+| `front`           | Function  | {{Section 2 of !VDAF}}|
+| `range`           | Function  | {{Section 2 of !VDAF}}|
+| `to_le_bytes`     | Function  | {{Section 2 of !VDAF}}|
+| `xor`             | Function  | {{Section 2 of !VDAF}}|
+| `zeros`           | Function  | {{Section 2 of !VDAF}}|
+{: #common-functions title="Common Functionalities from {{!VDAF}}."}
```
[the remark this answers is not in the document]
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Updated

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Revised the reference to the Appendix/Motivation section to make it clear. We have Pratik and David as contributors (line 45) -- I think this is OK. RE: NSF support, I think it makes sense to Ack if this was the case before.

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+
+Mastic also uses finite fields as specified in {{Section 6.1 of !VDAF}}. We
+usually denote a finite field by `F` and its Python class object, a subclass of
+`Field`, as `field: type[F]`.
+The following functionalities are as defined in {{Section 6.1 of !VDAF}}:
```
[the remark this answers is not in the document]
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+defined in {{Section 7.1 of !VDAF}}:
+
+| Functionality   | Type          | Definition                  |
+|:----------------|:--------------|:----------------------------|
+| `flp.decide`    | Class Method  | {{Section 7.1 of !VDAF}}    |
+| `flp.decode`    | Class Method  | {{Section 7.1.1 of !VDAF}}  |
+| `flp.encode`    | Class Method  | {{Section 7.1.1 of !VDAF}}  |
+| `flp.prove`     | Class Method  | {{Section 7.1 of !VDAF}}    |
+| `flp.query`     | Class Method  | {{Section 7.1 of !VDAF}}    |
+| `flp.truncate`  | Class Method  | {{Section 7.1.1 of !VDAF}}  |
```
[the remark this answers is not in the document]
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+| Functionality     | Type      | Definition            |
+|:------------------|:----------|:----------------------|
+| `byte`            | Function  | {{Section 2 of !VDAF}}|
+| `cast`            | Function  | {{Section 2 of !VDAF}}|
+| `concat`          | Function  | {{Section 2 of !VDAF}}|
+| `front`           | Function  | {{Section 2 of !VDAF}}|
+| `range`           | Function  | {{Section 2 of !VDAF}}|
+| `to_le_bytes`     | Function  | {{Section 2 of !VDAF}}|
+| `xor`             | Function  | {{Section 2 of !VDAF}}|
+| `zeros`           | Function  | {{Section 2 of !VDAF}}|
```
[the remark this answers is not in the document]
TrustworthyComputing: Removed `zeros`. We use `range`, `to_le_bytes` and `xor` in lines 460, 694, and 492 respectively.

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+|:----------------|:--------------|:----------------------------|
+| `flp.decide`    | Class Method  | {{Section 7.1 of !VDAF}}    |
+| `flp.decode`    | Class Method  | {{Section 7.1.1 of !VDAF}}  |
+| `flp.encode`    | Class Method  | {{Section 7.1.1 of !VDAF}}  |
+| `flp.prove`     | Class Method  | {{Section 7.1 of !VDAF}}    |
+| `flp.query`     | Class Method  | {{Section 7.1 of !VDAF}}    |
+| `flp.truncate`  | Class Method  | {{Section 7.1.1 of !VDAF}}  |
+{: #FLP-functionalities title="FLP Functionalities from {{!VDAF}}."}
+
+The following parameters are defined in {{Section 7.3.2 of !VDAF}}:
```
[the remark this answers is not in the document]
TrustworthyComputing: Updated

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+{: #circ-params title="Validity Circuit Parameters."}
+
+Mastic also uses eXtendable Output Functions (XOFs) as specified in
+{{Section 6.2 of !VDAF}}. The following functionalities are as defined herein:
+
+| Functionality       | Type          | Definition                  |
+|:--------------------|:--------------|:----------------------------|
+| `XofFixedKeyAes128` | Constructor   | {{Section 6.2 of !VDAF}}    |
+| `XofTurboShake128`  | Constructor   | {{Section 6.2 of !VDAF}}    |
+| `xof.next`          | Class Method  | {{Section 6.2 of !VDAF}}    |
```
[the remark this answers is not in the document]
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Fixed based on dst and dst_alg

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Updated

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+The following parameters are defined in {{Section 7.3.2 of !VDAF}}:
+
+| Parameter     | Type      | Definition                  |
+|:--------------|:----------|:----------------------------|
+| `MEAS_LEN`    | integer   | {{Section 7.3.2 of !VDAF}}  |
+| `OUTPUT_LEN`  | integer   | {{Section 7.3.2 of !VDAF}}  |
+{: #circ-params title="Validity Circuit Parameters."}
+
+Mastic also uses eXtendable Output Functions (XOFs) as specified in
+{{Section 6.2 of !VDAF}}. The following functionalities are as defined herein:
```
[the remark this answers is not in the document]
TrustworthyComputing: Fixed

draft-mouris-cfrg-mastic.md
```diff
@@ -216,11 +217,103 @@ into Mastic in {{vidpf}}. The appendix includes supplementary material:
...
+
+| Functionality       | Type          | Definition                  |
+|:--------------------|:--------------|:----------------------------|
+| `XofFixedKeyAes128` | Constructor   | {{Section 6.2 of !VDAF}}    |
+| `XofTurboShake128`  | Constructor   | {{Section 6.2 of !VDAF}}    |
+| `xof.next`          | Class Method  | {{Section 6.2 of !VDAF}}    |
+{: #XOF-functionalities title="XOF Functionalities from {{!VDAF}}."}
+
+> NOTE `xof` instances use a seed of size `SEED_SIZE` (in bytes).
```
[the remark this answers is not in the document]
TrustworthyComputing: Moved text to line 286

draft-mouris-cfrg-mastic.md
TrustworthyComputing: Revised description in Mastic

draft-mouris-cfrg-mastic.md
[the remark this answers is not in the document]
cjpatton: Excellent.

cjpatton (approved):
LGTM.

PR #90, opened 2024-10-19
Make aggregate API streaming

jimouris:
* Replace `aggregate()` with streaming API
* Update VDAF upstream to https://github.com/cfrg/draft-irtf-cfrg-vdaf/commit/8bbac2ea217ca4c980dceec9af6a99ffcc30deb3
* Update CODEOWNERS

Issue #72, opened 2024-09-17
Generate test vectors


cjpatton:
Define a few variants of Mastic, perhaps for Count, Sum, and Histogram. Add script for generating test vectors, similar to https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/poc/gen_test_vec.py, and use it to add test vectors to the repo.

PR #91, opened 2024-10-19
Add initial test vectors

jimouris:
Note: on top of https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/90 so review in order

Addresses: https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/72


poc/gen_test_vec.py
```diff
@@ -0,0 +1,109 @@
...
+        0,
+        (
+            mastic.vidpf.test_index_from_int(0b0, 1),
+            mastic.vidpf.test_index_from_int(0b1, 1),
+        ),
+        True,
+    )
+    gen_test_vec_for_vdaf(
+        vdaf_test_vec_path,
+        mastic,
```
cjpatton, line 33: I suspect mypy is complaining because it can't infer the types of generic parameters, i.e., `Measurement`, `AggResult`, and `F`. You probably need a subclass in which all the generic parameters are specified explicitly. Here's how we specify variants of Prio3:

https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/bd983637a87a7fa8353db5771c184585b2550271/poc/vdaf_poc/vdaf_prio3.py#L578-L590

In fact, we'll want to do something similar for Mastic so that we can at least specify an algorithm ID for each instance.

jimouris: Yeah, I tried different tricks but couldn't get past `mypy`. 

Ah, I see, good idea. I'll follow our examples and define:
* `MasticSum`
* `MasticCount`
* `MasticHistogram`
in `mastic.py`.

jimouris: Added both, in the same order as Prio3 does so we're consistent.

poc/mastic.py
```diff
@@ -21,7 +22,7 @@
 
 MasticAggParam: TypeAlias = tuple[
     int,                      # level
-    tuple[tuple[bool, ...]],  # prefixes
+    tuple[tuple[bool, ...], ...],  # prefixes
```
jimouris, line 25: @cjpatton double check this as I needed to change it for `mypy` to work.

poc/gen_test_vec.py
cjpatton, line 1: Let's add at least one variant that uses joint randomness, such as MasticSumVec.

cjpatton (commented):
Nice!

cjpatton (approved):

PR #96, opened 2024-11-08
Editorial fixes

divergentdave:
This makes a handful of editorial fixes in the spec and in comments.


jimouris (approved):
Thanks a lot, David!

PR #101, opened 2025-01-03
poc: Incorporate changes from Rust implementation

### LINKED ISSUE #80: Move joint randomness parts to input shares
Right now the public share encodes the FLP joint randomness parts, following the same lines as Prio3. However while working on the paper we decided to move these to the input shares to save some communication. In particular, we would add the helper's part to the leader's input share and the leader's part to the helper's input share.

### LINKED ISSUE #81: Avoid operating on serialized correction words
The FLP joint randomness computation binds to the VIDPF correction words and keys generated by the client. This means the aggregators have to either store the encoded correction words or re-encode them on the fly. Neither is desirable from an implementation perspective.

Two options:

1. Replace the encoded correction words in the binder with a hash of the correction words, which the aggregator could compute while decoding.
2. In @hannahdaviscrypto's implementation (https://github.com/divviup/libprio-rs/blob/dea0456ec4e0af8a3c1fed26db6bf7d776f195ad/src/flp/szk.rs#L459-L474), we bind only to the measurement shares. We suspect this is sufficient, and in fact it's implicit in the syntax of the paper.

### LINKED ISSUE #92: Idea: Compute the onehot and path consitency checks in one pass
At the moment we compute `onehot_proof` in the first traversal of the prefix tree, then compute `payload_check` in a second traversal. Then `onehot_proof || payload_check` is hashed to get `eval_proof`.

Both the `onehot_proof` and `payload_check` are made up of components, where each component corresponds to a node visited in the prefix tree. We could instead interleave these components, i.e., compute the hash of `node_proof_1 || payload_check_1 || node_proof_2 || payload_check_2 || ... || node_proof_i || payload_check_i || ...` where `i` is the index of the `i`th noted visited during traversal of the prefix tree. `node_proof` means the "node proof", that is, the proof that IDPF invariant holds for the node. (The onehotness proof is essentially the hash of the node proofs: see #79.)

cjpatton:
Closes #80.
Closes #81.
Closes #92.

We are working on a Rust implementation of Mastic: https://github.com/divviup/libprio-rs/issues/947

This commit incorporates a few improvements based on this experience:

1. Move the joint randomness part hints from the public share to the input shares. This simplifies the types and saves a bit of communication. (#80)

2. Let `beta` denote the value that is programmed into the output of the VIDPF. Currently the VIDPF prepends the counter internally.

3. Reorder the order in which the leader and helper seed are read from the sharding randomness. The new order ensures we always read the seeds in the same order regardless of whether FLP joint randomness is required.

4. When computing the FLP joint randomness, bind to the weight shares rather than to the VIDPF key share. (#81)

5. Add Mastic type parameters to the test vectors.

6. Swap the order of the joint rand part and the verifier share in the prep share encoding.

7. Prehash the payload check in order to avoid having to store the entire string. (#92)

8. VIDPF: Decouple onehot proof computation from `eval_next()`. We now compute the explicit prefix tree from the candidate prefixes, then compute both the onehot proof and payload check via a breadth-first traversal of the prefix tree.

9. VIDPF: Add a method for obtaining the aggregator's share of `beta`.


jimouris:
what do you mean in:
> Let beta denote the value that is programmed into the output of the VIDPF. Currently the VIDPF prepends the counter internally.

poc/tests/test_vidpf.py
```diff
@@ -22,13 +22,10 @@ def test_eval_invariants(self):
...
             PrefixTreeEntry.root(keys[1], True),
         ]
-        proof = [ONEHOT_PROOF_INIT, ONEHOT_PROOF_INIT]
         for i in range(vidpf.BITS):
             idx = PrefixTreeIndex(alpha[:i+1])
-            (node[0], proof[0]) = vidpf.eval_next(
-                node[0], proof[0], pub[i], ctx, nonce, idx)
-            (node[1], proof[1]) = vidpf.eval_next(
-                node[1], proof[1], pub[i], ctx, nonce, idx)
+            node[0] = vidpf.eval_next(node[0], pub[i], ctx, nonce, idx)
```
jimouris, line 27: nice

cjpatton:
> what do you mean in:
> 
> > Let beta denote the value that is programmed into the output of the VIDPF. Currently the VIDPF prepends the counter internally.

In the paper, we call "beta" the value that we program into the output during VIDPF key generation:
https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/main/poc/vidpf.py#L107

In our implementation, we prepend a "counter" (i.e., 1) to the beta internally in VIDPF:
https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/main/poc/vidpf.py#L196

However, in Mastic we need to consume this counter, so we end up breaking this abstraction. I think it's more natural to align with what we have in the paper.

poc/mastic.py
```diff
@@ -85,7 +81,7 @@ def __init__(self,
                  valid: Valid[W, R, F]):
         self.field = valid.field
         self.flp = FlpBBCGGI19(valid)
-        self.vidpf = Vidpf(valid.field, bits, valid.MEAS_LEN)
+        self.vidpf = Vidpf(valid.field, bits, 1 + valid.MEAS_LEN)
```
jimouris, line 84: I see, nice.

poc/mastic.py
```diff
@@ -108,74 +104,79 @@ def shard_without_joint_rand(
...
+        beta = [self.field(1)] + self.flp.encode(weight)
 
         # Generate VIDPF keys.
         (correction_words, keys) = \
             self.vidpf.gen(alpha, beta, ctx, nonce, vidpf_rand)
 
         # Generate FLP and split it into shares.
         prove_rand = self.prove_rand(ctx, prove_rand_seed)
-        proof = self.flp.prove(beta, prove_rand, [])
+        proof = self.flp.prove(beta[1:], prove_rand, [])
```
jimouris, line 122: so this assumes the counter is always first, right? Could you leave a comment somewhere to make this more explicit?

poc/mastic.py
```diff
@@ -108,74 +104,79 @@ def shard_without_joint_rand(
...
         (vidpf_rand, rand) = front(self.vidpf.RAND_SIZE, rand)
         (prove_rand_seed, rand) = front(self.xof.SEED_SIZE, rand)
-        (leader_seed, rand) = front(self.xof.SEED_SIZE, rand)
         (helper_seed, rand) = front(self.xof.SEED_SIZE, rand)
+        (leader_seed, rand) = front(self.xof.SEED_SIZE, rand)
         assert len(rand) == 0  # REMOVE ME
 
         (alpha, weight) = measurement
-        beta = self.flp.encode(weight)
+        beta = [self.field(1)] + self.flp.encode(weight)
```
jimouris, line 146: Maybe a comment here to explain it.

jimouris (approved):

poc/mastic.py
cjpatton: Done.

poc/mastic.py
cjpatton: Done.

PR #103, opened 2025-01-06
poc: Add additional test vectors

cjpatton:
Stacked on #101.

* MasticCount_2.json uses an aggregation parameter that is sensitive to the order in which the nodes are visited during the eval proof traversal.

* MasticCount_3.json uses an aggregation parameter that excludes the weight check.

### DIFF
--- a/poc/gen_test_vec.py
+++ b/poc/gen_test_vec.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 import os
 
 from vdaf_poc.test_utils import gen_test_vec_for_vdaf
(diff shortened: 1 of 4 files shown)


poc/gen_test_vec.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+# /usr/bin/env python3
```
cjpatton, line 1: This change was suggested by autopep. In fact, this file's permissions don't include exec, so we can probably just delete this line?

jimouris: oh, it knows more about formatting than I do :) either removing or just this change is fine.

jimouris (approved):

poc/gen_test_vec.py
cjpatton: Removed.

PR #104, opened 2025-01-07
poc: Move eval proof computation to Mastic

### LINKED ISSUE #100: VIDPF should be independent of Mastic
Specify VIDPF so that it can be used easily in an application other than Mastic. In particular:

1. The onehot, path, and counter checks should be moved into Mastic. VIDPF should just compute the prefix tree, including the weight and proof for each node.
2. The domain separation tag should not include the word "mastic".

cjpatton:
Closes #100.

Currently VIDPF computes the evaluation proof, which is largely Mastic-specific. For instance, we would like to bind Mastic preparation to the VDAF algorithm ID, as this allows for domain separation across different weight types. But other applications of VIDPF may have different requirements for domain separation.

This commit moves the eval proof computation to Mastic. VIDPF now just computes the prefix tree and returns the root and leaves it up to applications to do something with the result.

To maintain the current test coverage, we also make the following changes:

1. Add a method on VIDPF, `test_eval()` that computes the hash of the node proofs and returns that instead of the prefix tree.

2. Move the `test_malformed_correction_word_payload()` to the Mastic tests. The node proof is not actually sensitive to changes to the payload correction word, but payload check is.

### DIFF
--- a/poc/mastic.py
+++ b/poc/mastic.py
@@ -4,21 +4,29 @@ import itertools
 from typing import Any, Optional, TypeAlias, TypeVar, cast
 
 from vdaf_poc.common import (concat, front, to_be_bytes, to_le_bytes, vec_add,
-                             vec_sub, zeros)
+                             vec_sub, xor, zeros)
 from vdaf_poc.field import Field64, Field128, NttField
 from vdaf_poc.flp_bbcggi19 import (Count, FlpBBCGGI19, Histogram,
                                    MultihotCountVec, Sum, SumVec, Valid)
 from vdaf_poc.vdaf import Vdaf
 from vdaf_poc.xof import XofTurboShake128
 
-from dst import (USAGE_JOINT_RAND, USAGE_JOINT_RAND_PART,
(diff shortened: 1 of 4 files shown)


cjpatton:
Reviewer note: There are no functional changes in this PR. To verify, try `python gen_test_vec.py`. You'll notice that the test vectors don't change.

poc/mastic.py
```diff
@@ -253,6 +261,70 @@ def prep_init(
                 2,
             )
 
+        # Payload and onehot checks.
+        payload_check_binder = b''
+        onehot_proof = ONEHOT_PROOF_INIT
+        assert root.left_child is not None
+        assert root.right_child is not None
+        q = [root.left_child, root.right_child]
```
jimouris, line 269: what does `q` stand for?

jimouris (approved):

poc/mastic.py
cjpatton: "queue"

PR #106, opened 2025-01-07
poc: Use empty seeds instead of zero seeds

### LINKED ISSUE #102: Replace zero seeds with empty seeds
`XofTurboShake128` allows variable length seeds. So instead of using `zeros(32)` as the seed, as we've done so far in Prio3 and other places, we could use an empty seed, i.e., `b''`. This saves us from hashing a few bytes in the XOF input, which may save a Keccak evaluation in certain instances.

We currently use a mixture of zero and empty seeds. We should be consistent.

cjpatton:
Stacked on #105.
Closes #102.

`XofTurboShake128` is designed to tolerate variable length seeds. Use an empty seed, i.e., `'b''`, wherever a zero seed, `zeros(32)`, is currently used. This may save a Keccak evaluation in rare instances, especially wherever the binder string is long.


jimouris (approved):
nice

PR #107, opened 2025-01-07
poc: Simplify the onehot check

### LINKED ISSUE #79: Idea: Consider using Merkle-Damgard rather than the "linear" variant
Each call to `vidpf.eval_next()` updates the onehot proof as follows (excerpted from the spec):

```
# Compute and correct the node proof and update the onehot proof.
# Each update resembles a step of Merkle-Damgard compression. The
# main difference is that we XOR each block (i.e., corrected node
# proof) with the previous hash (or IV) rather than compress.
#
#             corrected node proof
#                 |
#                 |
#                 v
# current      +-----+     +------+     +-----+      updated
# proof  --+-->| XOR |---->| Hash |---->| XOR |----> proof
#          |   +-----+     +------+     +-----+
#          |                               ^
#          |                               |
#          +-------------------------------+
#
# [MST24]: \tilde\pi = H_1(x^{\leq i} || s^\i)
#          \pi = \tilde \pi \xor
#             H_2(\pi \xor (\tilde\pi \xor t^\i \cdot \cs^\i)
```

This is Merkle-Damgard-ish: the difference is that we XOR the next block (i.e., the node proof) with the current state (i.e., the current proof) then hash to get the next state, rather than compress the next block and the current state to get the next state. This comes from [CP22], but why do they do this? If we can get away with plain Merkle-Damgard, then it might be a good idea to use it here.

cjpatton:
Stacked on #106.
Closes #79.

For each node in the prefix tree that we traverse, we compute a "node proof" according to [CP22]. We then combine the node proofs into the "onehot proof" according to the following scheme.

Each node proof updates the onehot proof. An update resembles a step of Merkle-Damgard compression. The main difference is that we XOR each block (i.e., corrected node proof) with the previous hash (or IV) rather than compress:

```
            node proof
               |
               v
current     +-----+     +------+     +-----+    updated
proof --+-->| XOR |---->| Hash |---->| XOR |--> proof
        |   +-----+     +------+     +-----+
        |                               ^
        |                               |
        +-------------------------------+
```

This construction is also from [CP22]. According to the authors, any method of hashing the node proofs that is "secure against permutations attacks" should suffice. All we really want is to confirm that the node proofs computed by both servers are the same.

This commit replaces this bespoke construction with a single XOF invocation, using the node proofs in breadth first order as the binder.

This change results in fewer XOF usage domain separators, so renumber and reorder these separators as well.


poc/dst.py
```diff
@@ -17,22 +17,24 @@
...
-USAGE_EVAL_PROOF: int = 9
+# VIDPF
+USAGE_NODE_PROOF: int = 9
 USAGE_EXTEND: int = 10
 USAGE_CONVERT: int = 11
-USAGE_PAYLOAD_CHECK: int = 12
 
 
 def dst(ctx: bytes, usage: int) -> bytes:
+    assert usage in range(256)
```
jimouris, line 31: could you explain this?

poc/mastic.py
```diff
@@ -277,35 +271,21 @@ def prep_init(
...
+            # Update the onehot check.
+            onehot_check_binder += n.proof
 
         payload_check = self.xof(
             b'',
             dst_alg(ctx, USAGE_PAYLOAD_CHECK, self.ID),
             payload_check_binder,
         ).next(PROOF_SIZE)
 
+        onehot_check = self.xof(
```
jimouris, line 283: Are we calling it `check` from now on? Leaving a note to remember to update the .md

jimouris (approved):

poc/dst.py
cjpatton: `usage` should be a byte: I wanted to make sure the caller is actually passing a byte.

poc/mastic.py
cjpatton: Open for debate. I prefer "check" for consistency with "payload check" and "counter check".

poc/dst.py
jimouris: Shouldn't we check if it's <= than the biggest value we have (`11` here)?

poc/mastic.py
jimouris: fine by me.

cjpatton:
Rebased on main.

PR #108, opened 2025-01-07
poc: Make the evaluation proof pseudorandom

### LINKED ISSUE #73: Idea: Choices for `ONEHOT_PROOF_INIT`
`ONEHOT_PROOF_INIT` is used as the initial  proof, which is updated with node proof computed by each call to `eval_next()`. It appears we're free to set this to any public value:
* The original [de Castro and Polychroniadou paper](https://eprint.iacr.org/2021/580) sets this to the "correction seed" ($\mathsf{cs}$ in Algorithm 3).
* The PLASMA paper sets this to to "0" (the all zero string).
* We currently set it to a salt.

I had the thought that we could also derive this from the VDAF verification key so that it is pseudorandom from the point of view of the adversary (i.e., the malicious client). I wonder if this could provide some benefit in terms of security.

cjpatton:
Stacked on #107.
Closes #73.

Seed the XOF invocation with the VDAF verification key, thereby rendering the verification key pseudorandom from the point of view of the robustness attacker (i.e., malicious Clients). This doesn't appear to be necessary based on our security analysis so far, but it seems like a reasonable defense-in-depth measure.


jimouris (approved):

cjpatton:
Regenerated test vectors and rebased.

PR #109, opened 2025-01-07
poc: Let `beta` be the counter plus the encoded, truncated weight

### LINKED ISSUE #78: Idea: correction words for truncated weights
The FLP is run on the encoded measurement, which in general is longer than the truncated measurement. In some cases it might be much longer, e.g., for the `Sum` circuit, the encoded measurement is several field elements, while the truncated measurement is just one.

The encoded measurement is only needed when we verify the FLP. Suppose we modified Mastic as follows: 

1. Include shares of the encoded measurement, i.e., `flp.encode(weight)`,  in the input shares.
2. Instead of setting `beta` to be the encoded measurement share, set `beta` to be the truncated measurement, i.e., `beta = flp.truncate(flp.encode(weight))`.
3. Extend the path consistency check to make sure the left and right children of the root sum up to the encoded measurement.

cjpatton:
Stacked on #108.
Closes #78.

Depending on the FLP, truncation may significantly reduce the length of the measurement. For example, for the `Sum` type, the length of the encoded measurement depends on the validity range, but the truncated measurement is a single field element. We can therefore significantly reduce the size of the correction words by programming the VIDPF with the truncated weight instead of the full encoded weight.

To make this work, we secret share the encoded weight precisely as in Prio3. We then extend the payload check by setting the root payload to the counter plus the encoded truncated weight carried by the encoded input share.

This change only saves communication cost when `mastic.flp.MEAS_LEN > mastic.flp.OUTPUT_LEN` and `mastic.vidpf.BITS` is sufficiently large. Of the currently defined variants, only `MasticSum` will see savings. For `MasticCount`, `MasticSumVec`, and `MasticHistogram` the Leader's share increases by `mastic.flp.MEAS_LEN` field elements.

Another benefit of this change is that the payload check detects malformed correction word payloads at the first level of the tree.

Related clean up:

1. Add another usage domain separator for the helper's weight share and renumber.

2. Simplify the `MasticInputShare` type by defining classes for the leader and helper shares.

3. vidpf: Remove `get_beta_share()` as it's no longer used.


jimouris (approved):

cjpatton:
Fixed and regenerated test vectors, rebased.

cjpatton:
Closing due to https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/78#issuecomment-2593627448

PR #110, opened 2025-01-15
poc: Truncate the weight at each level but the first

### LINKED ISSUE #78: Idea: correction words for truncated weights

cjpatton:
Stacked on #111.
Closes #78.

First, extend VIDPF by adding an optional linear transform to the payload. The transform may depend on the level of the tree we're evaluating. Note that the length of each correction word depends on the length of the transformed payload.

Second, modify Mastic by applying the following transformation: for all but the first level of the tree, truncate the weight according to the FLP. We skip the first level so that we can still do the range check at the first level, but for each subsequent level, we don't need the full weight share.

Depending on the FLP, this may significantly reduce the size of the public share.


poc/mastic.py
```diff
@@ -305,15 +308,14 @@ def prep_init(
...
-        truncated_out_share = []
-        for val_share in out_share:
-            truncated_out_share += [val_share[0]] + \
-                self.flp.truncate(val_share[1:])
+        flattened_out_share = []
+        for out_share in out_shares:
+            # The VIDPF transform truncates at every level but the first.
+            if level == 0:
+                out_share = self.truncate(out_share)
+            flattened_out_share += out_share
```
jimouris, line 316: Why do we need to truncate the first level? Why not just pass `out_share` here?

jimouris, line 316: `flattened_out_share` --> `truncated_flattened_shares`?

poc/vidpf.py
```diff
@@ -81,6 +81,10 @@ def root(cls, seed: bytes, ctrl: bool):
         return cls(seed, ctrl, [], bytes())
 
 
+def identity_transform(w: list[F], _level: int) -> list[F]:
```
jimouris, line 84: Oh, that's nice. Initially I was about to comment why we need to abstract the truncate inside the transform in mastic, but I see why now, nice.

poc/mastic.py
cjpatton: Eh, this is a little subjective, but I think "truncated" is implied here.

jimouris: I got this part, what I'm missing is the why do we need to truncate? We already evaluated right, so why truncate after? (It might be obvious but I didn't get it from the description)

jimouris: Oh for the path checks! Right?

cjpatton: Are you asking why we need to truncate at all?

jimouris: I understand the truncation is to reduce the public share size - what I was missing is why we need to truncate the first level after we evaluate it.

poc/vidpf.py
```diff
@@ -361,7 +371,7 @@ def convert(self,
         xof = XofFixedKeyAes128(seed, dst(ctx, USAGE_CONVERT), nonce)
         next_seed = xof.next(XofFixedKeyAes128.SEED_SIZE)
         payload = xof.next_vec(self.field, self.VALUE_LEN)
-        return (next_seed, payload)
+        return (next_seed, self.transform(payload, level))
```
cjpatton, line 374: I'm wondering if it's a good idea to transform the derived payload. An alternative would be to compute the length of the transformed payload and simply generate that many field elements here.

cjpatton:
Re-generated test vectors.

cjpatton:
Rebased on #111.

jimouris (approved):

poc/vidpf.py
jimouris: I prefer it as it is now that we're calling transform.

hannahdaviscrypto (changes requested):

poc/mastic.py
```diff
@@ -305,15 +308,14 @@ def prep_init(
...
         ).next(PROOF_SIZE)
 
-        # Concatenate the output shares into one aggregatable output,
-        # applying the FLP truncation algorithm on each FLP measurement
-        # share.
-        truncated_out_share = []
-        for val_share in out_share:
-            truncated_out_share += [val_share[0]] + \
-                self.flp.truncate(val_share[1:])
+        flattened_out_share = []
```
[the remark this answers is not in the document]
cjpatton: It's not just renaming, it's also "flattening". That is, `out_shares` is a list of lists, and `flattened_out_share` is a list that consists of the out shares concatenated together.

poc/vidpf.py
cjpatton: That was my concern as well, though I think we're OK for the current set of Mastic variants, and probably any "natural" variant for which the output share is meaningfully secret, i.e, not just a constant.

The reason I didn't initially go with the length is that it adds a lot of code complexity to libprio.

cjpatton: (This change is much more invasive if have to deal with correction word decoding, which we don't here)

cjpatton:
Alright, I think I'm just gonna give up on this. This definitely works in theory, but the code design for libprio is turning out to be super hard. If desired we can revisit this in a future draft.

